PR #49, opened 2014-08-05
Separate API modules in jsforce

stomita:
Browserify build a script which includes all modules related, and it tends to be very large footprint even if user doesn't use all of the feature.

It should be separated into several modules under `jsforce` namespace, and allow to load partially like following:

``` html
<script src="/js/jsforce-core.min.js"></script>
<script src="/js/jsforce-api-streaming.min.js"></script>
<script src="/js/jsforce-api-metadata.min.js"></script>
```

### DIFF
--- a/lib/browser/request.js
+++ b/lib/browser/request.js
@@ -18,7 +18,7 @@ module.exports = function(params, callback) {
   var bufs = [];
   var sent = false;
   str._write = function(chunk, encoding, callback) {
-    bufs.push(chunk.toString(encoding));
+    bufs.push(chunk.toString(encoding === "buffer" ? "binary" : encoding));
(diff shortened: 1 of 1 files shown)


Krisa:
How is the core any different from the full lib? Both files on the cdn are actually exactly the same:
https://cdnjs.cloudflare.com/ajax/libs/jsforce/1.5.0/jsforce.min.js vs https://cdnjs.cloudflare.com/ajax/libs/jsforce/1.5.0/jsforce-core.min.js

Is there also a way to eventually reduce the overall payload, 300KB+ is quite big?

stomita:
This is the bug of build system and it would be resolved in 1.6.

PR #135, opened 2014-11-28
Adds system topic for streaming api, adds ability to set max sockets for request module

joeferraro:


Krisa:
@stomita  - Actually it is documented like we can see here: https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_apextestqueueitem.htm

Thanks to reconsider the merge, it would be very helpful to be able to write something like `conn.streaming.systemTopic('Logging').subscribe(listener);` or `conn.streaming.systemTopic('TestResult').subscribe(listener);`, etc.

Krisa:
Please ignore my request, I've just seen that it's already supported when the topic name starts with a slash https://github.com/jsforce/jsforce/blob/master/lib/api/streaming.js#L176

PR #130, opened 2015-01-05
Submitting batches via bulk.load when access key has expired errors out

stomita:
If I instantiate a Connection using a refresh token, and then use the Connection to bulk load records, I receive an error if the access token is invalid. 
When the access token has expired the returned HTTP status code is 400, not 401 as expected. This causes the Connection to return an error rather then trying to fetch another access token.


stomita:
I've mistakenly created pull request on this issue. This PR should be for #132.

PR #161, opened 2015-01-14
Update delete function

fourq:
Delete function fails due to http DELETE not being allowed. Changed to GET.


fourq:
Hold off on this. This is conditional on api version.

PR #193, opened 2015-03-13
Enable access to the SOAP API wrapper

doctorpangloss:
Seems to work great. Perhaps you want to expose access to the SOAP API for things like converting leads? That's what I'm using it for.

This is a super trivial change that I don't believe would qualify for a CLA, but I'd be happy to sign one :)

### DIFF
--- a/lib/jsforce.js
+++ b/lib/jsforce.js
@@ -9,3 +9,4 @@ exports.Connection = require('./connection');
 exports.OAuth2 = require('./oauth2');
 exports.Date = exports.SfDate = require("./date");
 exports.RecordStream = require('./record-stream');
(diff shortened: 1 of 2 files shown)


stomita:
The current SOAP class implementation is low level, and it will be used not only in Force.com SOAP API but in other APIs using SOAP protocol (e.g. Metadata API). The Force.com SOAP APIs like convertLead we are going to provide special interface to call in JSforce.

toooo:
@doctorpangloss Can u share code with how u implemented convertLead() using SOAP

THX!

stomita:
In #201 added SOAP interface

Issue #225, opened 2015-06-19
Add URL as property on query object


rfink:
Would be good to be able to access the url used to generate the query.  Pull request incoming.

stomita:
Not sure whether it should be merged to official master branch

PR #226, opened 2015-06-19
Adding url as property to query, test

rfink:
Addresses issue #225


stomita:
Not sure whether it should be merged to official master.

rfink:
Does this cause issues?  May I ask why it wouldn't make sense in the master?

stomita:
I don't think it is so evital for everyone, and as it is private it can be changeabl in future. You can patch in your folk.

PR #362, opened 2016-02-25
Added support for custom headers

bhaskarmelkani:
1. Added support for adding custom headers in GET/POST/PUT/PATCH/DELETE requests.
2. Have  backward compatibility support.
3. Updated package.json, to add index.js and core.js in files.


stomita:
Overloading body and settings might raise confliction. Consider the case apex method has parameters named 'headers' and 'body'.

bhaskarmelkani:
I can do it like this.
`Apex.prototype.get = function(path, body, callback, headers) ` so that headers is an optional field.

Suggest any better way.

stomita:
Aligning with other API interface in JSforce, adding optional `options` arg which can have `headers` property before `callback` arg would be preferable.

See http://jsforce.github.io/jsforce/doc/SObject.html#create

bhaskarmelkani:
Hey, I have updated the signature. Check if its fine ? :)

Issue #429, opened 2016-04-22
Handle multiple bulk query result seamleslly in a stream


stomita:
Fix [this](https://github.com/jsforce/jsforce/blob/1.6.0/lib/api/bulk.js#L801-L803)

Related issue: #428

tejasmanohar:
[stream-concat](https://www.npmjs.com/package/stream-concat) may help.

tejasmanohar:
Thanks!

PR #430, opened 2016-04-22
fix: combine all batches from bulk api into single stream

tejasmanohar:
I've done some basic testing on my end and am now using this via a fork for my project. Solves #429.

### DIFF
--- a/lib/api/bulk.js
+++ b/lib/api/bulk.js
@@ -15,7 +15,8 @@ var inherits     = require('inherits'),
     RecordStream = require('../record-stream'),
     CSV          = require('../csv'),
     Promise      = require('../promise'),
-    HttpApi      = require('../http-api');
+    HttpApi      = require('../http-api'),
+    joinStreams  = require('multistream');
 
 /*--------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


tejasmanohar:
@stomita Would you please review this? CI seems to be failing for unrelated cause. Let me know if there's anything I can do to help - Thanks!

stomita:
@tejasmanohar Thanks. It seems you use [multistream](https://www.npmjs.com/package/multistream) lib  , I'd like to know why you chose this lib ?

stomita:
@tejasmanohar I checked the [stream-concat](https://www.npmjs.com/package/stream-concat) you mentioned at first, and found that it has a feature of somewhat "lazy-opening" streams. It will save buffer memory instead of parallel fetching. How do you think of this approach ?

tejasmanohar:
@stomita It solves the problem we have- combining streams, and is recommended [here](https://github.com/felixge/node-combined-stream) by one of the former Node core contributors as an alternative to his module that supports the Node streams2 API. Plus, the implementation is concise & straight-forward, and it appears to be actively maintained. Most importantly, I feel comfortable debugging it in the case of problems. Do you see anything wrong with the library or have another suggestion? I'm open to whatever fits the use case.

tejasmanohar:
Also, I urge that this PR is released as a major change since the caveat of the feature was commented inline in the code (& maybe listed elsewhere? I could not find it in documentation)- plus, it introduces serious functionality changes that were never a part of this library. That said, I do suggest that we add a notice about it to the documentation because the prior functionality was not expected.

tejasmanohar:
Thanks! @stomita

Issue #428, opened 2016-04-21
Does Bulk Query stream include all batches?


tejasmanohar:
Does the Bulk Query `.stream()` include all batches? https://jsforce.github.io/document/#bulk-query

```js
var batch = conn.bulk.load("Account", "insert");
// MULTIPLE BATCHES -> ->
batch.on("response", function(rets) { // fired when batch finished and result retrieved
  for (var i=0; i < rets.length; i++) {
    if (rets[i].success) {
      console.log("#" + (i+1) + " loaded successfully, id = " + rets[i].id);
    } else {
      console.log("#" + (i+1) + " error occurred, message = " + rets[i].errors.join(', '));
    }
  }
);
//
// When input stream becomes available, pipe it to batch stream.
//
// -> -> ARE ALL INCLUDED HERE? -> ->
csvFileIn.pipe(batch.stream());
```

Also, is there a spec on what events `batch.stream()` emits?

tejasmanohar:
Also, if you use this style-
```js
conn.bulk.query("SELECT Id, Name, NumberOfEmployees FROM Account")
  .stream().pipe(fs.createWriteStream('./accounts.csv'));
```

How do you get the `.on('progress')` event to know when each poll request is sent out? The stream returned by `.stream()` doesn't seem to expose this.

stomita:
@tejasmanohar Unfortunately the current bulk query implementation (ver 1.6) doesn't handle multiple batch result (of course it should do).

tejasmanohar:
Hmmm so what happens if multiple results come back in the batch query response? is only the first one used? @stomita

tejasmanohar:
oh snap, that is the case ([ref](https://github.com/jsforce/jsforce/blob/master/lib/api/bulk.js#L803)). that should probably be in the docs given that it has the potential to drop data

tejasmanohar:
In what case are multiple batches returned? They can be returned at any time, right? Not specified by the user?

PR #433, opened 2016-04-25
fix: combine all batches from bulk api into single stream

### LINKED ISSUE #429: Handle multiple bulk query result seamleslly in a stream

stomita:
Close #429 #428


tejasmanohar:
Hmm does jsforce keep data from the stream in-memory somewhere? I seem to be getting OOMs after switching from my own handling to JSForce.

stomita:
It is assumed that the bulk query returns a record stream (= the stream contains record object) which parses the returned CSV only when any record stream consumer is connected. As current CSV  parser stores CSV string in memory the parsing may fail if the returned text is too large. 
If you are using record stream and get fail you can bypass the CSV parser by directly consuming retuned raw stream.

tejasmanohar:
When directly consuming the raw stream without the CSV parser (`.stream()`), it seems I'm facing OOMs still. May have some time to do a heapdump next week to see what's causing that, but it does not happen with my implementation of the bulk request stream -> csv parsing w/o jsforce.

tejasmanohar:
@stomita Would you please release this to NPM? Thanks!

PR #375, opened 2016-05-10
Connection#identity() returns XHR error in non-proxied environment

stomita:
From visualforce page it shows following error in conslole.

```
XMLHttpRequest cannot load https://login.salesforce.com/id/00D80000000L4RvEAK/00580000001hc9JAAQ?forma…

Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Origin 'https://mashmatrix-demo01-dev-ed--c.na6.visual.force.com' is therefore not allowed access.
```

It should use JSONP transport, but not used correctly.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -887,7 +887,10 @@ Connection.prototype.identity = function(callback) {
   ).then(function(res) {
     var url = res.identity;
     url += '?format=json&oauth_token=' + encodeURIComponent(self.accessToken);
-    return self.request(url, null, { jsonp : 'callback' });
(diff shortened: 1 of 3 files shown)


aguynamedben:
@stomita I'm not sure, but I believe this may have broken the identify() call for people that don't want to use JSONP. Would you mind taking a look at https://github.com/jsforce/jsforce/issues/611 and giving your feedback on what the best way to fix it is?

I'm new to jsforce but willing to do the work to make the change if you can advise on what the best option might be.

PR #479, opened 2016-05-23
Remove VERSION from core library.

paulmillr:
If you browserify the library, it would currently expose all sorts of private information; which is generated by NPM in `package.json`.

Telling users to do something to "avoid" the problem would not going to fix it. Insecure by default == insecure.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -7,8 +7,7 @@
 var EventEmitter = require('events').EventEmitter;
 
 var jsforce = module.exports = new EventEmitter();
-var pkg = require('../package.json');
-jsforce.VERSION = pkg.version;
+jsforce.VERSION = '';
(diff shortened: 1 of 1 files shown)


stomita:
@paulmillr  What type of private information is included in `package.json` ? Is there any information which are not publicly available from GitHub/NPM ?

paulmillr:
@stomita yes — it exposes full server path to the repo and similar.

paulmillr:
@stomita

stomita:
@paulmillr I'm not sure it does. The npm installed package.json only includes the information which are already available in NPM repos.

stomita:
@paulmillr Oh, npm 3 does it. Let me consider which is the best to deal with.

PR #489, opened 2016-06-12
Enable Execute Anonymous via SOAP api

celador:
@stomita Please review.
I would like to get this capability merged into your main fork.


lib/soap.js
```diff
@@ -120,7 +120,19 @@ SOAP.prototype.parseError = function(body) {
...
+    var newResponse = {
+        body: Body,
+        header: Header
+    };
+    this._conn.debuggingHeader = undefined;
+    return newResponse;
+  }
+  else {
+    return lookupValue(body, [ /:Envelope$/, /:Body$/, /.+/ ]);
+  }
```
stomita: Switching response type by the connection property is not acceptable because it will cause uncertain effects to the application code. If you want to debug the header information, just add debug log inside it.

PR #494, opened 2016-06-17
Enable to pass options in URL GET, identity, and retrieve calls to mo…

guanzhijie:
…dify http headers

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -333,7 +333,8 @@ Connection.prototype.request = function(request, options, callback) {
 Connection.prototype.requestGet = function(url, options, callback) {
   var request = {
     method: "GET",
-    url: url
+    url: url,
+    headers: options.headers
   };
   return this.request(request, options, callback);
 };
@@ -563,7 +564,7 @@ Connection.prototype.retrieve = function(type, ids, options, callback) {
     _.map(ids, function(id) {
       if (!id) { return Promise.reject(new Error('Invalid record ID. Specify valid record ID value')).thenCall(callback); }
       var url = [ self._baseUrl(), "sobjects", type, id ].join('/');
-      return self.request(url);
+      return self.request({ method: 'GET', url: url, headers: options.headers });
     })
   ).then(function(results) {
     return !isArray && _.isArray(results) ? results[0] : results;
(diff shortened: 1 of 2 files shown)


stomita:
Why don't you use headers in request ? You can pass headers even in GET request.

```js
conn.request({
  method: 'GET',
  url: url,
  headers: {
     'X-HTTP-HEADERS-YOU-WANT-TO-ADD': 'some value'
  }
}, function(err, res) {
  // ...
});
```

guanzhijie:
That's exactly what I did on line 299: when the 'request' is a single string (represents the url), I added headers from the options. I didn't change the request when the 'request' is an object. So whatever set in the request, including headers, can be used as it is. 

As for changes in identity and retrieve calls, please see my comments inline.

lib/connection.js
```diff
@@ -563,7 +565,7 @@ Connection.prototype.retrieve = function(type, ids, options, callback) {
     _.map(ids, function(id) {
       if (!id) { return Promise.reject(new Error('Invalid record ID. Specify valid record ID value')).thenCall(callback); }
       var url = [ self._baseUrl(), "sobjects", type, id ].join('/');
-      return self.request(url);
+      return self.request(url, options);
```
guanzhijie: Here I pass in 'options' to request so the headers in the options can be applied to the API call. Note 'url' here is a string, not an object.

I can also reconstruct the request as : 
return self.request({ method: 'GET', url: url, headers:  _.defaults(options.headers || {})});
But I'd like to keep info in 'options' encapsulated and let request to handle 'options', just in case there is any info other than headers can be used in request. If in the future we want to extend the usage of 'options' in request, the way I am using here will make it more flexible.

lib/connection.js
```diff
@@ -896,23 +898,29 @@ Connection.prototype.sobject = function(type) {
...
+    callback = options;
+    options = {};
+  }
   var self = this;
   var idUrl = this.userInfo && this.userInfo.url;
   return Promise.resolve(
     idUrl ?
     { identity: idUrl } :
-    this.request(this._baseUrl())
+    this.request(this._baseUrl(), options)
```
guanzhijie: Same case here and in line 923: url is a string, so I pass in options for request to retrieve headers and apply headers in API calls.

lib/connection.js
stomita: Just do it by 
```
this.request({ method: 'GET', url: url, headers: options.headers });
```
You don't have to extend `request()`, just change `retrieve()` to pass the headers info in options - which is already done in `create()` or `upsert()`.

stomita: > But I'd like to keep info in 'options' encapsulated and let request to handle 'options', just in case there is any info other than headers can be used in request.

https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

PR #496, opened 2016-06-18
Allow base64 output when retrieving metadata

Krisa:
specifying `conn.metadata.retrieve(...).stream('text')` allows to return straight the base64 instead of the binary data. That's wishful for instance when transferring the zip to the client instead of saving it on the server.


stomita:
If you want to get raw base64 string, why use `stream()` ? Just get promise/callback result and get zipFile property might be enough.

Krisa:
Right, by calling `complete`, instead of `stream`. This works, thanks.

PR #500, opened 2016-07-04
Use stream CSV parser library while handling bulkquery result CSV

### LINKED ISSUE #249: Streaming CSV Parser
Is there a plan to support streaming csv bulk queries to lessen impact on memory when the result set is more than a few hundred megabytes? 

It looks like the [current code](https://github.com/jsforce/jsforce/blob/master/lib/record-stream.js#L251) is downloading the entire CSV into memory, parsing it, and then emitting `record` events.

stomita:
Closes #249


tejasmanohar:
👏

PR #507, opened 2016-07-12
Use headers for requestGet, retrieve, identity, find and describe calls

guanzhijie:
This is to follow up the previous pull request https://github.com/jsforce/jsforce/pull/494. I've made more improvements based on @stomita comments, and set headers for more operations.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -333,7 +333,8 @@ Connection.prototype.request = function(request, options, callback) {
 Connection.prototype.requestGet = function(url, options, callback) {
   var request = {
     method: "GET",
-    url: url
+    url: url,
+    headers: options.headers
   };
   return this.request(request, options, callback);
 };
@@ -563,7 +564,7 @@ Connection.prototype.retrieve = function(type, ids, options, callback) {
     _.map(ids, function(id) {
       if (!id) { return Promise.reject(new Error('Invalid record ID. Specify valid record ID value')).thenCall(callback); }
       var url = [ self._baseUrl(), "sobjects", type, id ].join('/');
-      return self.request(url);
+      return self.request({ method: 'GET', url: url, headers: options.headers });
     })
   ).then(function(results) {
     return !isArray && _.isArray(results) ? results[0] : results;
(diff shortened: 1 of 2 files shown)


guanzhijie:
@stomita  can you please help check this pull request?

asifalimd:
I am actually looking for a way to send headers as part of get request. The need is I need to set the   header( **"Sforce-Query-Options": "batchSize=500"**) when I am using Query/QueryMore. your code seems to be fine but there is one place where the GET is called directly not from requestGet. I think the headers needs to be set there too.
Because when I debug the code while making query/querymore it seems the code on this method is executing.
in Connection.prototype.request
```
// if request is simple string, regard it as url in GET method
  if (_.isString(request)) {
    request = { method: 'GET', url: request };
  } 
```

guanzhijie:
@asifalimd I proposed that change in my previous pull request https://github.com/jsforce/jsforce/pull/494. See the conversation in that pull request, if I understand it correctly, I think @stomita prefers wrapping the headers in the request when making the call, rather than automatically adding the header into the request when the passed-in request is a url string in Connection.prototype.request.

stomita:
@asifalimd  In the request layer, the http header information is passed in parameter and there is no options argument. The `request(url: string)` is the short version of `request({ method: 'GET', url: url })` when you don't have to set http headers. The CRUD/business methods layer it has options info, and accepts the `options.headers` to pass down to request layer.

@guanzhijie I've added a few comments please check.

lib/sobject.js
```diff
@@ -530,12 +530,18 @@ ListView.prototype.results = function(callback) {
...
  */
-ListView.prototype.describe = function(callback) {
+ListView.prototype.describe = function(options, callback) {
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
   var url =  this._conn._baseUrl() + '/sobjects/' + this.type + '/listviews/' + this.id + '/describe';
-  return this._conn.request(url, callback);
+  return this._conn.request({ method: 'GET', url: url, headers: options.headers }, callback);
```
stomita: Fails when neither `options` and `callback` are given, which are optional args.

lib/connection.js
```diff
@@ -896,23 +897,29 @@ Connection.prototype.sobject = function(type) {
...
+    callback = options;
+    options = {};
+  }
   var self = this;
   var idUrl = this.userInfo && this.userInfo.url;
   return Promise.resolve(
     idUrl ?
     { identity: idUrl } :
-    this.request(this._baseUrl())
+    this.request({ method: 'GET', url: this._baseUrl(), headers: options.headers })
```
stomita: Fails when neither options and callback are given, which are optional args.

lib/connection.js
```diff
@@ -333,7 +333,8 @@ Connection.prototype.request = function(request, options, callback) {
 Connection.prototype.requestGet = function(url, options, callback) {
   var request = {
     method: "GET",
-    url: url
+    url: url,
+    headers: options.headers
```
stomita: As the `options` argument is for `HttpApi` configuration, the http headers info is not to be here.

lib/sobject.js
guanzhijie: Thanks @stomita  for catching it. I added a line as "options = options || {};"

lib/connection.js
guanzhijie: Thanks @stomita  for catching it. I added a line as "options = options || {};"

lib/connection.js
guanzhijie: Got it. Removed.

asifalimd:
@stomita, @guanzhijie , thanks for working on this. Adding the headers in options for HttpApi makes sense.

guanzhijie:
@stomita  any more feedback for this pull request?

tnoonan-salesforce:
@stomita Poke -

lib/connection.js
```diff
@@ -896,23 +896,30 @@ Connection.prototype.sobject = function(type) {
...
-    this.request(this._baseUrl())
+    this.request({ method: 'GET', url: this._baseUrl(), headers: options.headers })
   ).then(function(res) {
     var url = res.identity;
     url += '?format=json&oauth_token=' + encodeURIComponent(self.accessToken);
     var transport = Transport.JsonpTransport.supported ?
       new Transport.JsonpTransport('callback') :
       undefined;
-    return self.requestGet(url, { transport: transport });
+    return self.requestGet(url, options, { transport: transport });
```
stomita: The interface of `requestGet` is `requestGet(url, options, callback)`, options to be included in 2nd arg.

PR #516, opened 2016-08-09
Use headers for requestGet, retrieve, identity, find and describe calls

stomita:
#507


tnoonan-salesforce:
When would this show up in the npm?

Are you planning a release soon?

Tim

guanzhijie:
Thanks @stomita  for the merge and fix of request() options. Will you give a new release number for these changes so we can integrate soon?

PR #530, opened 2016-09-06
Added INCLUDES and EXCLUDES to the opMap

cthurston:
Multi-Select picklist data types use these keywords.


stomita:
@cthurston Could you please add some test cases in `soql-builder.js` to clarify the SOQL which new operator generates ?

PR #546, opened 2016-10-13
added platform events channel to streaming api

twojoke:
To prepare for this upcoming streaming feature: https://resources.docs.salesforce.com/204/latest/en-us/sfdc/pdf/platform_events_beta.pdf

### DIFF
--- a/lib/api/streaming.js
+++ b/lib/api/streaming.js
@@ -37,7 +37,7 @@ var Topic = function(streaming, name) {
  * @returns {Subscription} - Faye subscription object
  */
 Topic.prototype.subscribe = function(listener) {
-  return this._streaming.subscribe(this.name, "pushTopic", listener);
+  return this._streaming.subscribe(this.name, listener, "pushTopic");
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lib/api/streaming.js
```diff
@@ -169,25 +224,37 @@ Streaming.prototype.channel = function(channelId) {
...
  *
  * @param {String} name - Topic name
+ * @param {String} clientType - Client type
  * @param {Callback.<Streaming~StreamingMessage>} listener - Streaming message listener
  * @returns {Subscription} - Faye subscription object
  */
-Streaming.prototype.subscribe = function(name, listener) {
-  var channelName = name.indexOf('/') === 0 ? name : '/topic/' + name;
-  var fayeClient = this._getFayeClient(channelName);
+Streaming.prototype.subscribe = function(name, clientType, listener) {
```
stomita: This argument addition will raise incompatibility with old interface of `Stream#subscribe`.
We should not introduce such a hard as possible as we can.

lib/api/streaming.js
```diff
@@ -228,13 +228,19 @@ Streaming.prototype.channel = function(channelId) {
...
  * @returns {Subscription} - Faye subscription object
  */
-Streaming.prototype.subscribe = function(name, clientType, listener) {
-  var dir = "";
-  if (clientType === "pushTopic")
-    dir = "/topic/";
-  else if (clientType === "platformEvent")
-    dir = "/event/";
-  var channelName = dir + name;
+Streaming.prototype.subscribe = function(name, listener, clientType = "") {
```
stomita: Sorry we are not using ES6 so default parameter cannot be used here

Issue #566, opened 2016-11-15
Error from query payload


vinnyoodles:
https://github.com/jsforce/jsforce/blob/master/lib/query.js#L435

```javascript
TypeError: Cannot read property 'length' of undefined

 /var/app/current/node_modules/node_modules/jsforce/lib/query.js, line 435
```

PR #567, opened 2016-11-15
Check that data.records isn't undefined or null before checking the n…

ttacon:
…umber of records

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -432,7 +432,8 @@ Query.prototype._execute = function(options) {
     }
 
     // streaming record instances
-    for (var i=0, l=data.records.length; i<l; i++) {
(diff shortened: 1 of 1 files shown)


ttacon:
This is in reference to: #566

lib/query.js
```diff
@@ -432,7 +432,8 @@ Query.prototype._execute = function(options) {
     }
 
     // streaming record instances
-    for (var i=0, l=data.records.length; i<l; i++) {
+    var numRecords = (data.records && data.records.length) || 0;
+    for (var i=0, l=numRecords; i<l; i++) {
```
stomita: It would be enough
```javascript
for (var i=0, i<numRecords; i++) {
```

PR #568, opened 2016-11-22
Adding the ability to destroy reports via the analytics API

pcon:
I tried to follow as much prior art as possible.  I wasn't sure how to add a test for this so I didn't.  If you have any documentation on writing a test for it I'd be happy to do so.

### DIFF
--- a/lib/api/analytics.js
+++ b/lib/api/analytics.js
@@ -90,6 +90,23 @@ Report.prototype.destroy = function(callback) {
 };
 
 /**
+ * Clones a given report
+ *
+ * @method Analytics~Report#clone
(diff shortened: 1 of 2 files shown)


stomita:
@pcon You can setup your own test environment from here https://github.com/jsforce/jsforce/blob/master/test/SETUP.md
By the way, the destroying test is difficult to be done properly. You can implement both cloning and try destroying the cloned afterward.

pcon:
I've added the ability to clone a report and added testing for both the clone functionality and for the destroy functionality.

pcon:
@stomita do the included tests match your expectations?

PR #571, opened 2016-11-25
JSONP Transport 'supported' also checks for document

stanlemon:
In an environment such as React Native the ```window``` object is defined, but the ```document``` object is not, making the JSONP support not function correctly.  This PR adds a check to ```document``` as well. There shouldn't be any regressions from this change.


stanlemon:
Note, I am happy to rebuild the build dir after review.

PR #572, opened 2016-11-25
Switch to using webpack

stanlemon:
This PR switches to using webpack in place of browserify. The primary reason for this is that webpack does not utilize ```require()``` for it's loading calls, instead it creates functions named to itself.  This means that in hybrid environments like React Native where ```require()``` is defined, you will not run into errors trying to use JSForce.


stanlemon:
Note, I am happy to rebuild the build dir after review.

Issue #1, opened 2014-02-06
HTTP Proxy


rogah:
How can I setup a HTTP proxy to access the salesforce login url from within a company organization? I'm currently getting the following:

_Error: connect ETIMEDOUT_

Example:

```
var jsforce = require('jsforce');
var conn = new jsforce.Connection({
    proxy: 'http://proxy_ur:port'
});
```

jayfresh:
:+1: for releasing an update to npm :)

PR #581, opened 2016-12-10
Allow the ability to programmatically specify a proxy when using the library in node.

aotimme:
This implements the ability to specify a proxy programmatically when using jsforce on the server as a node module. This pull request piggybacks on the `proxyUrl` parameter and passes it through to the `request` module's `proxy` parameter. This also provides a fix for jsforce/jsforce#1.

The issue for us with using the `HTTP_PROXY` environment variable is that then all outgoing requests (at least using the `request` library) would be forced to use the proxy. This is because the `request` library also respects the `HTTP_PROXY` environment variable (https://github.com/request/request#controlling-proxy-behaviour-using-environment-variables). In our use case, we want to force all jsforce requests to go through a proxy but we don't necessarily want all other requests in that same node process to use the proxy.

Usage (as in jsforce/jsforce#1):
```
var jsforce = require('jsforce');
var conn = new jsforce.Connection({
    proxy: 'http://proxy_url:port'
});
```

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -43,7 +43,8 @@ var defaults = {
  * @param {String} [options.sessionId] - Salesforce session ID
  * @param {String} [options.refreshToken] - Salesforce OAuth2 refresh token
  * @param {String|Object} [options.signedRequest] - Salesforce Canvas signed request (Raw Base64 string, JSON string, or deserialized JSON)
- * @param {String} [options.proxyUrl] - In browser client, cross-domain proxy server URL, non Visualforce app; when in node, a proxy server URL
+ * @param {String} [options.proxyUrl] - Cross-domain proxy server URL, used in browser client, non Visualforce app.
+ * @param {String} [options.proxy] - URL of proxy server, used in server client.
  * @param {Object} [options.callOptions] - Call options used in each SOAP/REST API request. See manual.
  */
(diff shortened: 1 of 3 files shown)


aotimme:
Thanks for the feedback @stomita! Would you be more inclined to accept this if we create a new parameter `proxy` rather than overloading the `proxyUrl` parameter? If so I can make that change and you can review the updated pull request.

stomita:
@aotimme I prefer "httpProxy" rather than "proxy".

aotimme:
Thanks @stomita. Are you okay with the updated implementation otherwise? Specifically, are you okay with the new `ProxyServerTransport` class? I can call it `HttpProxyTransport` to be more in-line with using the term "httpProxy" if that sounds better.

aotimme:
I just changed it to use the parameter "httpProxy" so the usage looks like:
```
var jsforce = require('jsforce');
var conn = new jsforce.Connection({
    httpProxy: 'http://proxy_url:port'
});
```
Let me know if there are any other changes you'd like to see.

aotimme:
Thanks @stomita! Do you know when you're likely to issue a new release of this module (to npm) to include this? No worries if you don't have an estimate at this time.

jayfresh:
:+1: for releasing an update to npm

PR #586, opened 2016-12-23
Added support for layouts/recordTypeId requests

cthurston:


lib/sobject.js
```diff
@@ -413,15 +413,19 @@ SObject.prototype.deleted = function (start, end, callback) {
...
+SObject.prototype.layouts = function(layoutName, recordTypeId, callback) {
   if (typeof layoutName === 'function') {
     callback = layoutName;
     layoutName = null;
+  } else if(typeof recordTypeId === 'function'){
+    callback = recordTypeId;
+    recordTypeId = null;
   }
-  var url = "/sobjects/" + this.type + "/describe/" + (layoutName ? "namedLayouts/"+layoutName : "layouts");
+  var url = "/sobjects/" + this.type + "/describe/" + (layoutName ? "namedLayouts/"+layoutName : (recordTypeId ? "layouts/" + recordTypeId : "layouts"));
```
stomita: The function interface is for listing up the layout and should return Promise.<LayoutInfo> as callback. Specifying recordTypeId should be a different method for single LayoutInfo information to return.

Issue #597, opened 2017-02-14
JSForce is logging access token.


tnoonan-salesforce:
In version 1.7.1 of jsforce, specifically in connection.js. Salesforce access tokens are logged in the users' debug log. This is a potential security hole and makes salesforce vulnerable to impersonation.

PR #598, opened 2017-02-14
Remove access and refresh token logging.

tnoonan-salesforce:


tnoonan-salesforce:
https://github.com/jsforce/jsforce/issues/597

PR #606, opened 2017-02-23
Rewrite test to ES6+ and use AVA as test runner

stomita:
In order to moving much modern arch, rewrite the tests to ES6+ using async/await code.


stomita:
This pr was reverted because of the base brach is in invalid status

PR #627, opened 2017-04-26
Convert boolean inside JSON to string

vuza:
Currently I can upsert a single JSON containing boolean values like this: `conn.sobject('CustomObject').upsert({id: 1, flag: false})`.

However, I cannot use the same syntax for bulk operations, since `conn.sobject('CustomObject').bulkload('upsert', {}, [{id: 1, flag: false}])` leads SF to ignore the flag field.

To make bulk operations work, I had to stringify the boolean flag. `conn.sobject('CustomObject').bulkload('upsert', {}, [{id: 1, flag: 'false'}])` works as aspected.

Since the different APIs made me and my colleagues search for the error for quite a while, I fixed the bulk API by converting boolean values to strings.


lib/api/bulk.js
```diff
@@ -462,7 +462,14 @@ Batch.prototype.execute = function(input, callback) {
   } else {
     var data;
     if (_.isArray(input)) {
-      _.forEach(input, function(record) { self.write(record); });
+      _.forEach(input, function(record) {
+        self.write(record);
+        Object.keys(record).forEach(function(key) {
```
stomita: I think this part should come before `self.write()`

vuza: True! Something happened here, tx. Fixed that

amakk:
I'm having a similar issue when running `conn.sobject('Custom_Object__c').find({ Id: { $in: ['objectid'] } }).update({ Flag__c: false })`. `Flag__c` isn't getting updated when boolean value is false. When it's true, or a string it does get updated.

PR #635, opened 2017-05-01
Rewrite test to ES6+ and use AVA as test runner (cont.)

stomita:


stomita:
Mistekenly merged to master. reverted.

Issue #658, opened 2017-06-27
Expose status code as part of the error object


ccnixon:
Thanks so much for your excellent work. I was wondering if there was a way to pass the statusCode from the http response as part of the returned error object. Likely an additional property added here:

https://github.com/jsforce/jsforce/blob/master/lib/http-api.js#L251

Is it possible to access this currently?

Thanks!

cristiand391:
https://github.com/jsforce/jsforce/issues/490#issuecomment-2127711394

Issue #490, opened 2016-06-13
Error status code. 


danrc:
Is it possible to get error http status code?

vuza:
I was looking for the same feature. Unfortunately the http status code gets removed from the error object in this function, except Salesforce did not return a error in the response body: https://github.com/jsforce/jsforce/blob/a3ce63633f92672edcb536bbc22b6afcf2f592b2/lib/http-api.js#L250

cristiand391:
triaging old issues...

you can see the request status code by setting the log level to debug when creating the connection:
https://jsforce.github.io/jsforce/types/connection.ConnectionConfig.html
https://github.com/jsforce/jsforce/blob/af8a10e84557d95e20a62dae37e9dbf1e4d50fa8/src/http-api.ts#L120

some weird errors might include the status code, IIRC this was added to catch API returning HTML:
https://github.com/jsforce/jsforce/blob/af8a10e84557d95e20a62dae37e9dbf1e4d50fa8/src/http-api.ts#L312C1-L320C13

PR #659, opened 2017-06-28
Add statusCode to error object

tcigrand:
Closes https://github.com/jsforce/jsforce/issues/658


borys-kupar:
Is there any chance this could be merged? The HTTP status code would be very helpful, so it would be possible to distinguish type of failure, and make retry request based on that.

Is there any help needed? e.g. Opening new PR against latest master?

Thanks!

gpsamson:
Hey all! I'm a bit late to the party, is there a chance we could merge this in @stomita? We retry all Salesforce behavior because there's no way to see if the original error is due to a 400 or 500 response group.

Thanks!

promerSFDC:
I'm curious why this was never merged? The lack of an http status code is really annoying. Handing error responses differently based on the code returned seems like something most folks would want/could use.

cristiand391:
hey, sorry for the late reply, closing this since v1 is deprecated, see this comment about how to find the HTTP status code for all requests:
https://github.com/jsforce/jsforce/issues/490#issuecomment-2127711394

PR #662, opened 2017-07-11
Implementation of authentication for proxy server

vergara:
Changes required on the client side if proxy has authentication enabled. Older clients will continue working if authentication is disabled in the proxy.


vergara:
@stomita Please review.

stomita:
@vergara What is the use case of this functionality ? Is there any implementation that accepts authentication in ajax proxy ?

PR #668, opened 2017-07-21
add  DuplicateRuleHeader with SOAP API

tee-tee-dev:
Sometimes we want to insert records, without duplicate rules errors.
we added DuplicateRuleHeader with SOAP API .


tee-tee-dev:
The callOptions only set Header.callOptions. 
and DuplicateRuleHeader  is **not supported** with REST API . 
We want to set Header.DuplicateRuleHeader  for SOAP API.

PR #670, opened 2017-07-21
Added support for revoking API access by passing in the refresh token

lilawren:
As per [this reference page](https://developer.salesforce.com/blogs/developer-relations/2011/11/revoking-oauth-2-0-access-tokens-and-refresh-tokens.html)

"...revoking a refresh token terminates an app’s authorization to call APIs on behalf of the user and revokes all outstanding access tokens issued against that refresh token"

By passing in the refresh token, Salesforce admins won't have to revoke the token manually in Manage Apps => Connected Apps OAuth Usage


lib/connection.js
```diff
@@ -1101,14 +1101,16 @@ Connection.prototype.loginBySoap = function(username, password, callback) {
 /**
  * Logout the current session
  *
+ * Setting revoke to true will revoke API access
+ *
  * @param {Callback.<undefined>} [callback] - Callback function
  * @returns {Promise.<undefined>}
  */
-Connection.prototype.logout = function(callback) {
+Connection.prototype.logout = function(callback, revoke = false) {
```
stomita: The argument order should set callback to last.

lilawren: Updated in 9699c90

lib/connection.js
```diff
@@ -1099,28 +1099,28 @@ Connection.prototype.loginBySoap = function(username, password, callback) {
...
- *
- * Setting revoke to true will revoke API access
+ * Logout the current session 
  *
+ * @param {Boolean} revoke - Revokes API Access if set to true
  * @param {Callback.<undefined>} [callback] - Callback function
  * @returns {Promise.<undefined>}
  */
-Connection.prototype.logout = function(callback, revoke = false) {
+Connection.prototype.logout = function(revoke = false, callback) {
```
stomita: The change will break current API client which uses `Connection#logout(callback)` without revoke parameter.
Please check the first parameter is function or not, like `Connection#request()` does for the options.
https://github.com/jsforce/jsforce/blob/132f5823178d34524feed4fde075f31ad4751b96/lib/connection.js#L294-L298
Also you cannot use default parameter in ES5.

lilawren: Good catch, updated in d125c3d

Issue #547, opened 2016-10-17
replay (streaming api)


twojoke:
Any plans on implementing message durability using Replay, where we can pass a replay ID to the client so we can pull all messages that occurred from that ID forward?

For reference: https://developer.salesforce.com/docs/atlas.en-us.api_streaming.meta/api_streaming/using_streaming_api_durability.htm

stomita:
(The PR #546 is for this enhancement issue I understand)

stomita:
Actually it has already supported the replay channel, but not documented well.

sth-fish:
no support for replay?

jatkinson-vivial:
If replay is already supported how is the replay id set?

In the streaming class the _createClient method sets the endpoint but not the replayId

nolson-berry:
I believe this has been resolved now with: https://github.com/jsforce/jsforce/pull/740

PR #676, opened 2017-08-08
Added the ability to pass an extension to the faye client when subscri…

jatkinson-vivial:
…bing to a topic. Thus allowing client to pass in an extension with replay values.
Issue: https://github.com/jsforce/jsforce/issues/547


lib/api/streaming.js
```diff
@@ -127,12 +139,17 @@ Streaming.prototype._createClient = function(replay) {
...
   var isGeneric = channelName.indexOf('/u/') === 0;
   var clientType = isGeneric ? 'generic' : 'pushTopic';
   if (!this._fayeClients || !this._fayeClients[clientType]) {
     this._fayeClients = this._fayeClients || {};
     this._fayeClients[clientType] = this._createClient(isGeneric);
+
+    if (typeof extension !== "undefined") {
+      this._fayeClients[clientType].addExtension(extension)
+    }
+
```
stomita: Implicitly modifies faye client with addExtension, which leads the following normal getFayeClient(channelName) calls to return the client instance with extension.
Should be much explicit API interface if you are going to modify the client behavior.

sfescape: I have the opposite view, I think extensions should just be an added optional parameter when creating a Topic. The faye client is specific to the Topic so adding extensions to it should be treated the same as the channel name.

sfescape:
Why not just add extensions as the last (optional) argument to Topic?

conn.streaming.topic(name, extensions)

That seems the most natural place to put this. Not sure why it hasn't been accepted yet.

nolson-berry:
@stomita 
We just upgraded to the latest jsforce and the changes from @mars are working nicely.

**This PR can be closed.**  The _jatkinson-vivial_ github account is no longer active.

```javascript
const jsforce = require('jsforce')
...
function loginCallback (salesforceConnection) {
  const topic = `/topic/${process.env.SALESFORCE_TOPIC}`
  const replayExt = new jsforce.StreamingExtension.Replay(topic, -2)
  const fayeClient = salesforceConnection.streaming.createClient([replayExt])

  fayeClient.subscribe(topic, (salesforceEvent) => {
    console.log('Received: %j',salesforceEvent)
    return doSomething(salesforceEvent)
  })
}
...
```

PR #689, opened 2017-09-20
Adding support to the Bulk module to be able to insert binary data (i…

gla5001:
…e. Attachments).


.gitignore
```diff
@@ -5,10 +5,14 @@ lib-cov
 *.out
```
gla5001: Opening this PR to add support for different content types in the Bulk module to be able to insert binary data in bulk. Realized it wasnt supported, and I really kind of need it

lib/api/bulk.js
```diff
@@ -32,13 +32,16 @@ var inherits     = require('inherits'),
  * @param {Object} [options] - Options for bulk loading operation
  * @param {String} [options.extIdField] - External ID field name (used when upsert operation).
  * @param {String} [options.concurrencyMode] - 'Serial' or 'Parallel'. Defaults to Parallel.
+ * @param {String} [options.contentType] - The content type of the data.
+ *  Possible values 'CSV', 'XML', 'JSON', 'ZIP_CSV', 'ZIP_XML', or 'ZIP_JSON'. Defaults to 'CSV'.
  * @param {String} [jobId] - Job ID (if already available)
```
gla5001: All the possible values the contentType can be according to Salesforce documentation

lib/api/bulk.js
```diff
@@ -374,12 +377,36 @@ Batch.prototype._createRequestStream = function() {
...
+      break;
+    case 'ZIP_XML':
+      contentType = 'zip/xml';
+      break;
+    case 'ZIP_JSON':
+      contentType = 'zip/json';
+      break;
+    default:
+      contentType = 'text/csv';
+  }
```
gla5001: make sure to set contentType correctly when we create the request steam

package.json
```diff
@@ -72,6 +72,7 @@
     "xml2js": "^0.4.16"
   },
   "devDependencies": {
```
gla5001: added some helpful libs to make testing this easier

test/bulk.test.js
```diff
@@ -102,6 +105,110 @@ describe("bulk", function() {
   /**
    *
    */
```
gla5001: here are the tests. Basically it takes 1 record from the bulk insert above and inserts some attachments on it. It copies over a requestTemplate.txt file into the directory to be zipped, replaces a regex with a valid Id, zips it up, and sends the request off as a stream of the zip

lib/api/bulk.js
bdmayes: You might consider throwing these values into an enum somewhere. That would also afford you to the opportunity to verify that the given option is a valid one, and throw an error if it's not.

lib/api/bulk.js
```diff
@@ -374,12 +377,36 @@ Batch.prototype._createRequestStream = function() {
   var batch = this;
   var bulk = batch._bulk;
   var logger = bulk._logger;
+  var contentType;
+
+  switch(batch.job.options.contentType) {
```
bdmayes: You could also compare to enum values here in the switch. And if you're feeling really feisty, you could stack the CSV case right above the default without a break, so that it falls through into the default.

.gitignore
phillipthurmond: @stomita @xgeek-net @tnoonan-salesforce Hey guys! Any idea when the next release of jsforce might be and whether you would consider this feature? it's pretty important for us. Thanks!

stomita: @phillipthurmond I'm now on the way of releasing 2.0, which is rewritten ES2017 + Flow, and freezing current 1.8.x for accepting additional features. I cannot say the releasing date of 2.0 as it has several issues I found already. The PR will be reviewed again after the release of 2.0.

cristiand391:
Hey, sorry this didn't get reviewed in time, we aren't doing feature work on master and there were some refactors in the bulk module in v3 which may require work to get this work again.

I've left the GH discussion open in case someone have use-cases for this feature, we can re-consider then:
https://github.com/jsforce/jsforce/discussions/1505

PR #695, opened 2017-10-21
return Response Header when caller set returnResponseHeader to true i…

jennifersalesforce:
@stomita 
Would you please review my pull request?
I want to return request headers for query if option.returnResponseHeader is true.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -207,7 +207,7 @@ Connection.prototype.initialize = function(options) {
   }
 
   this.limitInfo = {};
-
+  this.responseHeaders = {};
   this.sobjects = {};
   this.cache.clear();
   this.cache.get('describeGlobal').on('value', _.bind(function(res) {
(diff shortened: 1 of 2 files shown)


lib/connection.js
```diff
@@ -310,16 +310,21 @@ Connection.prototype.request = function(request, options, callback) {
...
+            if (apiUsage) {
+                self.limitInfo = {
+                    apiUsage: {
+                        used: parseInt(apiUsage[1], 10),
+                        limit: parseInt(apiUsage[2], 10)
+                    }
+                };
+            }
+        }
+        if (options.returnResponseHeaders) {
```
jennifersalesforce: return responseHeaders when options.returnResponseHeaders is true.

jennifersalesforce:
@stomita 
Would you please review my pull request and let me know if I need to change anything?

Thanks 
Jennifer

jennifersalesforce:
@stomita @joeferraro @stanlemon @icecubed 
Would you please review my change to see if I can merge?

Thanks
Jennifer

tnoonan-salesforce:
@stomita Poke.

@jennifersalesforce Maybe add @stomita to the reviewer list.

jennifersalesforce:
@stomita 
Looks like I don't have permission to add reviewers.
Would you please help to review it?

lib/connection.js
amphro: I think this needs to be reworked. First, `self` is the connection, not the request. So you could be dealing with a response that is not from your request. Second, why do we need to set an option to get the response headers? Instead, I would like to see this as part of the httpApi that is passed with the callback.

stomita:
@jennifersalesforce 
I'm against for storing response headers in connection properties, it is not the proper place.

Issue #697, opened 2017-10-27
BUG (with fix): RecordStream.Parsable Cannot Stream More Than ~1M Results from Bulk.query


andymjames:
Hi,

This is a wall of text, and I know it's a lot, but I am merely trying to distill all of the debugging I've been doing the last couple of days, and convince you that I've done due diligence and have a pretty good understanding of what's happening and why. I hope that ultimately the details provided will give you the confidence that I have indeed found a bug, and an appropriate fix.

Best,
Andy

Fix:
1) Remove the call to this.push(record) in https://jsforce.github.io/jsforce/doc/record-stream.js.html#line38
2) Remove the highWaterMark: 500 * 1000 setting in https://jsforce.github.io/jsforce/doc/record-stream.js.html#line123 (this change isn't required, but you'll see below that I think the current setting is a workaround that will be obsolete after you make change 1)

Bug Report:

I was trying to Bulk.query() the ID of every record for an Object with 1.8M records. In doing so, I believe I have found a bug in the implementation of RecordStream.Parsable that makes it impossible to Bulk query more than 1,000,016 records (even though the server is able to provide them).

On line 38 in the _transform method of RecordStream (https://jsforce.github.io/jsforce/doc/record-stream.js.html#line38) you call this.push(record); and I've found in my testing that after 1,000,016 calls to it, that it returns false indicating that the buffer is full and can't take anymore pushes, which then causes the stream to silently error and it stops emitting records and never emits the "end" event. The records seem to hit my code as a result of the this.emit('record', record); call on the line above, which casts doubt about the necessity of the call to push.

I have a feeling that you hit this somewhere in your coding/testing, because it seems you tried to do a workaround on line 123 in the Parsable.stream method (https://jsforce.github.io/jsforce/doc/record-stream.js.html#line123) by boosting the highWaterMark to 500k Objects. When I was first trying to play with your code, my first hacky workaround was to double your value for highWaterMark to be 1 million, and sure enough my 1.8 million records came back just fine. I've experimented and the number of records your current code will return is equal to 16 + 2 * highWaterMark setting on line 123. I don't know why it's double the value, but assume the extra 16 is from the default highWaterMark on the dataStream declared on 121.

[...]

PR #698, opened 2017-10-27
Fix Bug Limiting Number of Records RecordStream.Parsable Can Handle

andymjames:
More details in the issue I filed here: https://github.com/jsforce/jsforce/issues/697

This change removes the call to this.push(record) in the _transform method of RecordStream, which causes some buffer that isn't getting read to overflow after 1,000,016 pushes, which manifests itself such that anyone using Bulk.query() to fetch more than 1M16 records has their stream cut off prematurely.

I have also removed the highWaterMark setting later on, as I found it was unnecessary once I removed the call to this.push() above, which caused me to assume that overriding the default there was a workaround for the bug in _transform, and it simply took a long time for anyone to exceed the limit of the workaround that was in place.

### DIFF
--- a/lib/record-stream.js
+++ b/lib/record-stream.js
@@ -34,7 +34,6 @@ inherits(RecordStream, Transform);
  */
 RecordStream.prototype._transform = function(record, enc, callback) {
   this.emit('record', record);
-  this.push(record);
   callback();
 };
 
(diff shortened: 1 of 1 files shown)


stomita:
@andymjames if you remove `this.push(record)` it will not be passed to downstream. It shows failure in the bulk insert test.

andymjames:
And I'm talking about failures in _fetching_ records. So this makes it pretty clear to me that you cannot have RecordStream.Parsable AND RecordStream.Serializable both inherit from RecordStream as is. I mentioned this as a caveat in the issue I linked.

Or you need to detect which direction the stream is going and only call .push for data going to SFDC and only call record.emit for data being queried.

You most definitely have a bug. If you don't believe me, then load 1,000,017 records into an object and try querying for all of them... it _will_ fail after 1,000,016.

My fix may not be sophisticated enough to prevent RecordStream.Serializable from failing tests, but I think that means that RecordStream.Serializable and RecordStream.Parseable need less shared inheritance.

Issue #704, opened 2017-11-30
coffee-script has moved to coffeescript


tomanagle:
Hi, 

The package coffee-script has moved to coffeescript. 
Ref: https://www.npmjs.com/package/coffee-script

This is causing a warning on Ubuntu and an error on Windows.

> > coffee-script@1.12.8 postinstall /var/www/my_product/node_modules/coffee-script
> > node --eval 'if (require("./package.json").name === "coffee-script") { var red, yellow, cyan, reset; red = yellow = cyan = reset = ""; if (!process.env.NODE_DISABLE_COLORS) { red = "\x1b[31m"; yellow = "\x1b[33m"; cyan = "\x1b[36m"; reset = "\x1b[0m"; } console.warn(red + "CoffeeScript has moved!" + reset + " Please update references to " + yellow + "\"coffee-script\"" + reset + " to use " + yellow + "\"coffeescript\"" + reset + " (no hyphen) instead."); console.warn("Also, a new major version has been released under the " + yellow + "coffeescript" + reset + " name on NPM. This new release targets modern JavaScript, with minimal breaking changes. Learn more at " + cyan + "http://coffeescript.org" + reset + "."); console.warn(""); }'
> 
> CoffeeScript has moved! Please update references to "coffee-script" to use "coffeescript" (no hyphen) instead.
> Also, a new major version has been released under the coffeescript name on NPM. This new release targets modern JavaScript, with minimal breaking changes. Learn more at http://coffeescript.org.
>

clayzermk1:
Still an issue in jsforce 1.8.0.
> npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)

```
<snip>
├─┬ jsforce@1.8.0
│ ├── coffee-script@1.12.7
<snip>
```

Coooi:
Facing same issue here.
ERR! fetch failed http://npm.rz.is/coffeescript/-/coffeescript-1.12.8.tgz

Deepaksgithub:
npm WARN deprecated coffee-script@1.8.0: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)

cristiand391:
This should be fixed in v3, see: https://github.com/jsforce/jsforce/discussions/1529

PR #705, opened 2017-11-30
Update package.json

zenpc:
CoffeeScript on NPM has moved to coffeescript (no hyphen). Please update references to coffee-script to use coffeescript instead.


thvd:
Fixes: #704

tomanagle:
Do we know when this will be merged with master?

peternhale:
This change will not prevent the error on Windows. If the eval is causing the npm install to fail, just changing from `coffee-script` to `coffeescript` will not prevent the eval from failing. If version 1 of coffee script is needed for jsforce, then fixing coffee script version to `1.12.7` might be the best choice.

PR #731, opened 2018-03-13
Updating browser client to use js-cookie

aheuermann:
Setting document.cookie in the popup window is flaky for me in current chrome on OSX.

When I call `jsforce.browser.login` it runs [`removeTokens`](https://github.com/jsforce/jsforce/blob/master/lib/browser/client.js#L133) which [unsets the cookie](https://github.com/jsforce/jsforce/blob/master/lib/browser/client.js#L215). It then sets the cookie on successful login within the popup window. I then refresh the opener and the value is still blank. I noticed after [setting the cookie=true](https://github.com/jsforce/jsforce/blob/master/lib/browser/client.js#L204) in the popup it contains the same key twice in the developer console. Using js-cookie resolves this issue for me.

### DIFF
--- a/lib/browser/client.js
+++ b/lib/browser/client.js
@@ -10,7 +10,8 @@ var events = require('events'),
     qs = require('querystring'),
     _ = require('lodash/core'),
     Connection = require('../connection'),
-    OAuth2 = require('../oauth2');
+    OAuth2 = require('../oauth2'),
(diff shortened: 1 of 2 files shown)


cristiand391:

If someone experience issues with cookies on modern browsers please file a new issue.

PR #736, opened 2018-03-29
Refresh accessToken when salesforce returns 403 Bad Oauth Token

krashidov:
I'm doing a request to get the user's indentity information like so: 

sf.requestGet(`${instanceUrl}/id/${organizationId}/${userId}`)

The problem is that when the accessToken expires, this endpoint returns a 403, which I guess is different than the rest of salesforce because the current jsforce codebase only refreshes on a 401. The proposed PR forces a refresh of the accessToken on 401 or 403.


stomita:
This change will force token refresh when the record or resource is not accessible by the ACL, which should not be refreshed. I'm not sure it should be cared in this lib (seems like identity URL matter), but if we take care of it, it should be scoped to the identity URL only.

krashidov:
@stomita Thanks for the reply that makes sense. I have no problem with manually refreshing when we do the identity call, but it doesn't seem like there's a public method to force a refreshToken call? If there is, I'd be happy to use it.

I updated the PR to scope the 403 check to the login url. Not sure if that causes the same issues.

lib/http-api.js
```diff
@@ -205,8 +205,8 @@ function parseText(str) { return str; }
  * Detect session expiry
  * @protected
  */
-HttpApi.prototype.isSessionExpired = function(response) {
-  return response.statusCode === 401;
+HttpApi.prototype.isSessionExpired = function(response, request) {
+  return response.statusCode === 401 || (response.statusCode === 403 && request.url.indexOf('login') > -1);
```
stomita: Cannot be determined it is login server URL (identity URL) even it includes a "login" word

PR #746, opened 2018-04-20
Use user's responeType while sending xhr

Varun-garg:
We have already documented `responseType` parameter in `options` while sending a request, however it is not actually set into xhr.
This is useful while defining responseType to something like a blob.

Eg, when we are downloading an attachment on browser, we can skip conversion process and straightaway get url of blob.  

    var urlCreator = window.URL || window.webkitURL;
    record.Body = ''; // Attachment url received 
    var options = { responseType: 'blob' };
    connection.request(record.Body, options, function (err, blob) {
        if (err) { return console.error(err); }

        var location = urlCreator.createObjectURL(blob); // do something like set img src
        console.log("location: " + location);
    }

**References**
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType

Signed-off-by: Varun Garg <varun.10@live.com>

### DIFF
--- a/lib/browser/request.js
+++ b/lib/browser/request.js
@@ -12,6 +12,12 @@ module.exports = function(params, callback) {
     }
   }
   xhr.setRequestHeader("Accept", "*/*");
+
+  // Respect user's responseType to xhr
+  if(params.responseType != null && params.responseType) {
+    xhr.responseType = params.responseType;
+  }
+
   var response;
   var str = new Duplex();
   str._read = function(size) {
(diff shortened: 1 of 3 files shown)


stomita:
The `responseType` property used in jsforce's `HttpApi#request()` is not equal to the XHR's property. It accepts mimeType like "application/xml" , but XHR requires "document", "blob" or etc.
It is used to parse the body correctly in the client.
https://github.com/jsforce/jsforce/blob/7a8336368d45a390ed44f39daf60d6d598a95195/lib/http-api.js#L137
https://github.com/jsforce/jsforce/blob/7a8336368d45a390ed44f39daf60d6d598a95195/lib/http-api.js#L145

Issue #751, opened 2018-05-16
jsforce relies on vulnerable version of `open` dependency


zmillman:
The Node Security Platform flagged `open` as a critically vulnerable library: https://nodesecurity.io/advisories/663

It's currently [used for making requests from the CLI](https://github.com/jsforce/jsforce/blob/38e4e66d021db20b22435f48c9d2b34f48e54c23/lib/cli/cli.js#L11), and could probably be replaced by a normal `http` request.

In the meantime, this is causing any services depending on `jsforce` to fail their security checks.

jaan:
Even `sshpk` package is highly vulnerable. Posting the `npm audit` result below:

┌───────────────┬──────────────────────────────────────────────────────────────┐
│ **High**          │ Regular Expression Denial of Service          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ **Package**       │ sshpk          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ jsforce-metadata-tools [dev]          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ jsforce-metadata-tools > jsforce > request > http-signature  │
│               │ > sshpk          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/606          │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ **Critical**      │ Command Injection          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ **Package**       │ open          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ jsforce-metadata-tools [dev]          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ jsforce-metadata-tools > jsforce > open          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/663          │
└───────────────┴──────────────────────────────

PR #752, opened 2018-05-18
replace deprecated `open` with `opn`

vkmita:
Fixes: https://github.com/jsforce/jsforce/issues/751, https://nodesecurity.io/advisories/663

`open` appears to have a command injection vulnerability. This is causing any project with `jsforce` as a dependency to fail the node security platform check https://github.com/nodesecurity/nsp. 

The fix is to use `opn`.  Luckily the API is the same, so very little to change.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -8,7 +8,7 @@
 
 var http = require('http'),
     url = require('url'),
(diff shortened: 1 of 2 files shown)


danambrose:
Can we get this pushed through, please?

PR #757, opened 2018-05-31
Allow send XML or raw data with Apex

LiborVilimekMassive:
I need to make this request:

POST /services/apexrest/KeyDate

``` 
<?xml version="1.0" encoding="UTF-8"?>
<callout>
    <parameter name="AccountID">{Zuora account Id}</parameter>
</callout>
```

If JSON.stringify is forced and content-type as well, I cannot do this. However with the update above, this call will go through:

```
    const body = `<?xml version="1.0" encoding="UTF-8"?>
    <callout>
        <parameter name="AccountID">${subscription.accountId}</parameter>
     </callout>`;

    return connection.apex.post('/KeyDate/', body, { headers: { 'Content-Type': 'application/xml' } });
```

### DIFF
--- a/lib/api/apex.js
+++ b/lib/api/apex.js
@@ -43,7 +43,7 @@ Apex.prototype._createRequestParams = function(method, path, body, options) {
   if (body) {
     params.body = body;
     if (typeof body === 'object')
-     params.body = JSON.stringify(body);
+      params.body = JSON.stringify(body);
     }
   }
   return params;
--- a/lib/api/apex.js
+++ b/lib/api/apex.js
@@ -1,51 +1,177 @@
+/**
+ * @file Manages Salesforce Apex REST endpoint calls
+ * @author Shinichi Tomita <shinichi.tomita@gmail.com>
+ */
 
-import * as _ from 'lodash';
+'use strict';
 
-import {
-    State,
-    Subscription,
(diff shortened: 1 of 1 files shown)


LiborVilimekMassive:
PS: I would prefer to have some fields in options like "sendBodyRaw, dontChangeContentType", but not sure if you want to add fields like that

lib/api/apex.js
```diff
@@ -36,12 +36,15 @@ Apex.prototype._createRequestParams = function(method, path, body, options) {
...
   }
-  if (!/^(GET|DELETE)$/i.test(method)) {
+  if (!/^(GET|DELETE)$/i.test(method) && !_headers["Content-Type"]) {
     _headers["Content-Type"] = "application/json";
   }
   params.headers = _headers;
   if (body) {
-    params.body = JSON.stringify(body);
+    params.body = body;
+    if ('object' === typeof options['body']) {
```
stomita: Maybe
```javascript
if (typeof body === 'object')
```

LiborVilimekMassive: @stomita - thanks, fixed

LiborVilimekMassive:
@stomita - fixed

cristiand391:
closing this b/c v1 is now deprecated

PR #760, opened 2018-06-05
Replace the open dependency with opn

WhiteAbeLincoln:
There is currently an advisory for [open](https://github.com/pwnall/node-open): https://nodesecurity.io/advisories/663.
This is not a huge issue, since it doesn't appear that node-open ever gets passed unsanitized user input. However users of the jsforce library do see a critical severity vulnerability because of node-open when running `npm audit`. This pull request removes the node-open dependency and replaces it with [opn](https://github.com/sindresorhus/opn), which is more secure and actually maintained.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -8,7 +8,7 @@
 
 var http = require('http'),
     url = require('url'),
-    openUrl = require('open'),
+    openUrl = require('opn'),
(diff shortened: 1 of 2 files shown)


WhiteAbeLincoln:
Haha, I didn't check the open pull requests before creating this. #752 beat me. Closing

PR #769, opened 2018-06-26
Get instanceUrl from signedRequest

curiosity26:
When authorizing via a signed request, the instanceUrl is may not be set in the options passed to the initializer. At any rate, the instanceUrl should be set to the value found on the signed request, as this is the authenticating service.


cristiand391:
Hey, sorry this didn't get reviewed in time, we aren't feature work on master so I'm going to close this.

PR #773, opened 2018-07-03
Add additional CRUD functions using "composite resources", to better handle writes to a "medium" number of records

davidknoll:
When `create` or `update` are called with many records, these are broken down into individual requests. Therefore there are a lot of requests in flight at once. This has caused issues for us including delays and `UNABLE_TO_LOCK_ROW` errors.

In some places in our app, we have used the Bulk API instead, but this is not ideal where an interactive/immediate response is expected. It can lead to timeouts where an error is returned to our app, but the bulk action eventually goes ahead successfully.

The [Composite Resources](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite.htm), specifically the "SObject Tree" or "SObject Collections", sound like a good solution to this, as we get to submit a number of records at once and get a quick response.

I have therefore added `createMany`, `updateMany` and `destroyMany` functions (with synonyms) to the `Connection` and `SObject` (but not `RecordReference`) classes. These take the same parameters and return the same data as the original `create`, `update` and `destroy`, but use the composite resource underneath. I have not implemented a `retrieveMany` or `upsertMany`.

Using the "SObject Collections" part of "Composite Resources" required increasing the API version to 42. I have included tests for my new functions. I have also included a tweak to a permission set in the testsuite.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -689,6 +689,8 @@ Connection.prototype.createMany = function(type, records, options, callback) {
   records = isArray ? records : [ records ];
   if (records.length > self.maxRequest) {
     return Promise.reject(new Error("Exceeded max limit of concurrent call")).thenCall(callback);
+  } else if (!records.length) {
+    return Promise.resolve([]).thenCall(callback);
   }
   records = _.map(records, function(record) {
     var sobjectType = type || (record.attributes && record.attributes.type) || record.type;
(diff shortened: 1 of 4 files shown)


stomita:
Thanks, I'm supposing this feature (=composite resource) should replace the current normal CRUD calls which consumes a REST API for each records in parallel, not by the dedicated `xxxMany` call. In 1.9 this feature will be merged.

davidknoll:
I've added them as separate functions, but it should be fine if you use them to replace the originals, including for single records. Need to be careful of edge/error cases, and the requirement for API 42.0 as a minimum, and that I didn't implement the full set of functions. Also my versions still obey the maxRequest setting (appears to be 10 by default, but the underlying endpoint can handle up to 200).

Thanks!

PR #781, opened 2018-07-25
Use SObject collection API for query update/destroy operation

stomita:
As far as the queried record num is not so big as the given threshold, use collection API instead of bulk API, which is a little heavy and has always delay of polling.
The threshold value can be modified by passing options.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -629,7 +629,7 @@ Query.prototype.filter = RecordStream.prototype.map;
  * @method Query#delete
  * @param {String} [type] - SObject type. Required for SOQL based query object.
  * @param {Callback.<Array.<RecordResult>>} [callback] - Callback function
- * @returns {Bulk~Batch}
+ * @returns {Promise.<Array.<RecordResult>>}
  */
 /**
  * Synonym of Query#destroy()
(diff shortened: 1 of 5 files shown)


lib/query.js
```diff
@@ -623,85 +623,174 @@ Query.prototype.map = RecordStream.prototype.map;
...
  * @param {Record|RecordMapFunction} mapping - Mapping record or record mapping function
  * @param {String} [type] - SObject type. Required for SOQL based query object.
+ * @param {Object} [options] - Mass update operation options
+ * @param {Boolean} [options.allowBulk] - Allow switching to Bulk API when the num of queried records reached to certain threshold. Default is true.
+ * @param {Number} [options.bulkThreshold] - Threshold num to switch to use Bulk API instead of usual `SObject#delete()` call. Default value is 200 after API ver 42.0, and 0.5 * `maxRequest` before API ver 42.0.
  * @param {Callback.<Array.<RecordResult>>} [callback] - Callback function
  * @returns {Promise.<Array.<RecordResult>>}
  */
-Query.prototype.update = function(mapping, type, callback) {
+Query.prototype.update = function(mapping, type, options, callback) {
```
wctej89: should function handle case if `options` is a `function`?

PR #785, opened 2018-07-25
remove JSONP endpoint as it is not working in modern browsers

stomita:
In Chrome, CORB (Cross-Origin Read Blocking) prevents to load JSONP response from identity url / revoke server

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1201,11 +1201,7 @@ Connection.prototype.identity = function(options, callback) {
     this.request({ method: 'GET', url: this._baseUrl(), headers: options.headers })
   ).then(function(res) {
     var url = res.identity;
-    url += '?format=json&oauth_token=' + encodeURIComponent(self.accessToken);
-    var transport = Transport.JsonpTransport.supported ?
-      new Transport.JsonpTransport('callback') :
-      undefined;
-    return self.request({ method: 'GET', url: url }, { transport: transport });
+    return self.request({ method: 'GET', url: url });
   }).then(function(res) {
     self.userInfo = {
       id: res.user_id,
--- a/lib/oauth2.js
+++ b/lib/oauth2.js
@@ -157,24 +157,14 @@ _.extend(OAuth2.prototype, /** @lends OAuth2.prototype **/ {
    * @returns {Promise.<undefined>}
    */
   revokeToken : function(token, callback) {
-    var req;
-    if (Transport.JsonpTransport.supported) {
-      var jsonpTransport = new Transport.JsonpTransport('callback');
-      req = jsonpTransport.httpRequest({
-        method: 'GET',
-        url : this.revokeServiceUrl + '?' + querystring.stringify({ token: token })
-      });
-    } else {
-      req = this._transport.httpRequest({
-        method : 'POST',
-        url : this.revokeServiceUrl,
-        body: querystring.stringify({ token: token }),
-        headers: {
-          "Content-Type": "application/x-www-form-urlencoded"
-        }
-      });
-    }
-    return req.then(function(response) {
+    return this._transport.httpRequest({
+      method : 'POST',
+      url : this.revokeServiceUrl,
+      body: querystring.stringify({ token: token }),
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded"
+      }
+    }).then(function(response) {
       if (response.statusCode >= 400) {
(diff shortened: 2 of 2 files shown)


rynop:
@stomita I still am unable to use the `identity()` method after your fix (I'm using v1.9.0).

Your fix causes an `OPTIONS` CORS pre-flight request to be sent (good) but it gets denied by SFDC because it does not have the `Authorization` header present on the request.

For example, with your fix, this pre-flight `OPTIONS` request is sent:
```
curl -v 'https://test.salesforce.com/id/00D1F000000DQEfUAO/0051F000001XFwIQAW' -X OPTIONS -H 'Access-Control-Request-Method: GET' -H 'Origin: https://me.rynop.dev:5500' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36' -H 'Access-Control-Request-Headers: authorization'
```

If you run this command in the terminal you will see SFDC servers return a `403` with a `Missing_OAuth_Token` response body

If you send the `Authorization` header on the request, it works from the terminal, HOWEVER it wont work in the browser, because the response does not have the correct `Access-Control-Allow-Origin` response header.

Ex:
```
curl -v 'https://test.salesforce.com/id/00D1F000000DQEfUAO/0051F000001XFwIQAW' -X OPTIONS -H 'Access-Control-Request-Method: GET' -H 'Origin: https://me.rynop.dev:5500' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36' -H 'Access-Control-Request-Headers: authorization' -H 'Authorization: Bearer MyOAuth2Token'
```

As you probably know, XHRs (and fetch for that matter) [can't send](https://stackoverflow.com/questions/40722700/add-authentication-to-options-request) `Authorization` headers in the pre-flight `OPTIONS` request.  SO - I really think this is a bug in SalesForce - however I'm a total SFDC noob, so want your opinion 1st.  Thoughts?

I've never opened a bug against SFDC - got any contacts internally?  This is a pretty big deal IMO because you can't get user info without a proxy (YUCK).  Even the (should-be) standardized OAuth2 endpoint doesn't allow CORS: 
```
curl -v 'https://test.salesforce.com/services/oauth2/userinfo' -X OPTIONS -H 'Access-Control-Request-Method: GET' -H 'Origin: https://acmeloans.loanos.dev:5500' -H 'Referer: https://me.rynop.dev:5500/' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36' -H 'Access-Control-Request-Headers: authorization' --compressed
```

stomita:
@rynop Essentially the SFDC should support the CORS for identity URLs, but as you mentioned currently not. For that reason we did change the transport to JSONP for identity method only, but the w/a is not working right now in Chrome and also will not be in other modern browsers.

We have following two options to send request to SFDC:
1. Support CORS for Identity URLs (Access-Control-Allow-Origin: *)
2. Return correct response header (application/javascript) with JSONP response to pass CORB check

The first option is the best way IMO but needs several steps (IdeaExchange vote or something). The second one seems easy because they already support JSONP interface, but not nowadays.

Considering the needs of JSforce in browser is mostly in Visualforce, the Ajax proxy service is always available in Salesforce. Using proxy when using identity call might be the practical solution.

rynop:
@stomita thanks for the quick response.  Was afraid you'd say that (use proxy) :)

I'm super new to SFDC, but my guess is its like turning the titanic to get something fixed/implemented ([this basic building block](https://success.salesforce.com/ideaView?id=08730000000BroxAAC) is 11 years old).

I've opened [an idea](https://success.salesforce.com/ideaView?id=0873A000000PZE9QAO) on IdeaExchange but whats the most successful way to resolve an issue? Create a [support case](https://help.salesforce.com/articleView?id=workcom_contact_support.htm&type=5)?  I'm going to DreamForce for the 1st time this year, is there avenues to raise issues there?

Totally un-related: any idea how to edit an ideaexchange idea that you posted? It formatted like crap (didnt honor my bulleted lists that shows up in the WYSIWYG editor).

PR #803, opened 2018-09-07
Correct invalid CRUD save error handling function name

### LINKED ISSUE #801: Bug: _toSaveResult error
I'm seeing an issue with some methods in the `Connection` class where `_toSaveResult` is not found. In my particular case, I was using the `upsert` method and forgot to pass in the `extIdField`. The API was, rightfully, returning an error but when it went to run the `catch` block, it referenced the _toSaveResult` method which is nowhere to be found. Is this a dead reference?

```
connection.js:987
.catch(function(err) {
        // be aware that `allOrNone` option in upsert method will not revert the other successful requests
        // it only raises error when met at least one failed request.
        if (!isArray || options.allOrNone || !err.errorCode) { throw err; }
        return this._toSaveResult(null, err);
      })
```

stomita:
Fixes #801

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -660,7 +660,7 @@ Connection.prototype._retrieveMany = function(type, ids, options) {
  * @typedef RecordResult
  * @prop {Boolean} success - The result is succeessful or not
  * @prop {String} [id] - Record ID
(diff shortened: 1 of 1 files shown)


vlamic:
When this is planned to be released? Thanks!

PR #822, opened 2018-10-09
Added ability to call the metadata API describeValueType

paustint:
Added support for describeValueType which has been supported since API ver 33
Added unit tests for new method

https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_describeValueType.htm


cristiand391:
closing this now that v1 is deprecated

Issue #867, opened 2019-02-22
Ability to pass `rejectUnauthorized: false` to request to allow self-signed SSL certificates for certain Connection instances


aguynamedben:
Thanks for the work you do on this library, we like it a lot. 🙏

I'm building an Electron app that connects to Salesforce. Certain customers see the following error when jsforce tries to call the Salesforce API:

```
self signed certificate in certificate chain
```
I believe this is because these employees are behind a corporate firewall that is terminating SSL with a corporate self-signed SSL certificate (i.e. basically a [corporately approved MITM "attack" on employees for Salesforce traffic](https://stackoverflow.com/questions/39639860/dealing-with-corporate-firewall-adding-self-signed-certificates)). When this case is detected, ideally I'd show some UI to the user that says "Allow self-signed certificate when calling Salesforce?", they'd click "Allow", and our app would re-create the jsforce `Connection` instance with the underlying `request` instance having `strictSSL: false` (or `rejectUnauthorized: false`).

I see [in the jsforce code](https://github.com/jsforce/jsforce/blob/master/lib/transport.js#L15) that I can set `request.defaults`, but I don't want to do this for EVERY `request` object in my app, just certain jsforce `Connection` instances. Ideally `Connection` would allow for `options.requestLibOptions` and I could set `options.requestLibOptions.strictSSL = false` if I wanted to.

Is that somehow do-able with the current jsforce library? From my read of the code, it'd require a pull request to allow custom options for the `request` library. Probably a branch [around here](https://github.com/jsforce/jsforce/blob/master/lib/connection.js#L81) to detect for `options.requestLibOptions` and pass them to `new Transport()`?

There's another hacky option to set `process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;` for the entire process, but that's heavy-handed and unnecessary if I can just tweak how jsforce is using request.

I'm happy to submit a PR to allow `request` to be configured per-Connection if that makes sense and there are no current workarounds.

cristiand391:
triaging old issues...

jsforce v2 moved from `request` to `node-fetch` which doesn't seem to have a similar option like `strictFalse`.
Also, we are trying to minimize custom changes to the current node-fetch wrapper as we aren't sure we'll keep in future jsforce versions.

PR #868, opened 2019-02-22
`options.requestLibOptions` to customize calls made with request lib

aguynamedben:
Allow an option when creating a Connection,
`options.requestLibOptions`, which will be the basis for each call
made with the underlying
[request](https://github.com/request/request#requestoptions-callback)
module. Example usage:

```js
    const client = new jsforce.Connection({
      oauth2: {
        clientId: process.env.SALESFORCE_CONSUMER_KEY,
      },
      instanceUrl: tokens.instanceUrl,
      accessToken: tokens.accessToken,
      requestLibOptions: {
        strictSSL: false, // Allow self-signed certificates
        timeout: 1, // Set timeout to 1ms
      },
    });
```

Related GitHub issue: https://github.com/jsforce/jsforce/issues/867


aguynamedben:
Do I need to do anything to get this to work in the browser? I tested that it works in Node.js, and noticed this comment:

```
 * @param {String} [options.httpProxy] - URL of HTTP proxy server, used in server client.
```

I'm guess I should add a similar note "used in server client", right?

cristiand391:
https://github.com/jsforce/jsforce/issues/867

PR #913, opened 2019-07-25
add describeSOjbects endpoint to soap API

ori-moisis:
Added the describeSObjects API to soap (useful for bulk queries)


willy2dg:
It would be great if this were added.

Issue #925, opened 2019-09-19
Vulnerability: csv-parse - Regular Expression Denial of Service


buptliuhs:
```
                       === npm audit security report ===

# Run  npm install csv-parse@4.4.6  to resolve 1 vulnerability
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ csv-parse                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ csv-parse                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ csv-parse                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1171                            │
└───────────────┴──────────────────────────────────────────────────────────────┘


found 1 high severity vulnerability in 9460 scanned packages
  1 vulnerability requires semver-major dependency updates.
```

apasrini:
Even i am getting this issue csv parse needs to be updated to version 4.4.6 or higher 
https://www.npmjs.com/advisories/1171

jruizvaldiviaffdc:
I have the same issue, any idea when the module will be updated to use csv-parse 4.4.6 or higher?

createthis:
@stomita what is the plan here?

acrosman:
I am working on a simple PR to bump this to the current version. While NPM flag this as a breaking change since the current locked version is 1.3.3, I don't actually see any use in the JSForce code base that would actually break since the calls are pass through as best I can tell.  

However, the changes _could_ impact downstream users if they are using CSV-Parse in ways that have changed.

apasrini:
@stomita Could you please check on this issue as it is turning high priority vulnerability issue for us. I see a PR from @acrosman Thanks for this.

acrosman:
Fixed by #930

keanedawg:
Should this be closed?

buptliuhs:
Fixed in 1.9.3

PR #930, opened 2019-10-11
Bump csv-parse to 4.6.3

acrosman:
This addresses the security issue reported by NPM and discussed in #925.

It does not appear that JSForce itself needs to change, however in theory this could be a breaking change for downstream projects since it is a change in major version of csv-parse and calls passing through may be impacted if they are using deprecated calls.

[csv-parse changelog](https://csv.js.org/parse/changelog/) Obviously reports lots of changes since that time including dropping support for old versions of Node, move to the MIT license, and more aggressive skipping of empty records.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2284,9 +2284,12 @@
       }
     },
     "csv-parse": {
-      "version": "1.3.3",
-      "resolved": "https://registry.npmjs.org/csv-parse/-/csv-parse-1.3.3.tgz",
-      "integrity": "sha1-0c/YdDwvhJoKuy/VRNtWaV0ZpJA="
+      "version": "4.6.3",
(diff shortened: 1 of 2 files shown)


tnoonan-salesforce:
@stomita What's the chances of getting this into a patch build of jsforce this week?

dileepkandyala:
I have the same issue. Does this commit merged ? I still don't see a change when I did npm install jsforce.

Please let me know how can I update the csv-parser to resolve my issue.

tnoonan-salesforce:
@stomita Poke.

stomita:
It seems the latest csv-parse has non-ES5 code (class statement), and it causes failure in building libs for browsers. There is no transpilers (like babel) in this version of JSforce, it has to avoid the usage of the code. Fortunately it has the version of es5 by `require('csv-parse/lib/es5')`. I'll check if it works

PR #955, opened 2019-12-08
Added create method to Analytics API

ajgriesemer:
I've been working on a project to automate migration of reports from an existing org to a new org. In the process I stumbled into a report create method in the Analytics API . This patch adds a create method.


cristiand391:
closing b/c v1 is deprecated.

For reference, this can be done via `conn.request`:
https://jsforce.github.io/document/#request

Issue #813, opened 2018-09-27
REST API - upsert on collections with { allOrNone: false } does not return array of errors


maciejsimm:
I'm using upsert with option `allOrNone: true`, I'd expect that in return I'd receive an array of objects simillar to `Database.SaveResult[]`.

In case some record fails in the collection I'm getting a JSForce exception instead:

```
TypeError: Cannot read property '_toRecordResult' of undefined
    at /Users/maciejsimm/Documents/Workspaces/Salesforce/Enxoo/enxooProductMigrator/node_modules/jsforce/lib/connection.js:992:21
    at tryCallOne (/Users/maciejsimm/Documents/Workspaces/Salesforce/Enxoo/enxooProductMigrator/node_modules/promise/lib/core.js:37:12)
    at /Users/maciejsimm/Documents/Workspaces/Salesforce/Enxoo/enxooProductMigrator/node_modules/promise/lib/core.js:123:15
    at flush (/Users/maciejsimm/Documents/Workspaces/Salesforce/Enxoo/enxooProductMigrator/node_modules/asap/raw.js:50:29)
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
    at process._tickCallback (internal/process/next_tick.js:181:9)
```

I've checked and it's not the case for insert operation - which properly returns array of errors that can be individually handled.

My code is below (`data` is an array of sObjects). I'd like to upsert records, and in case individual ones have error -> log it to console, but still allow the upsert on correct records.

```
conn.sobject("enxCPQ__Attribute__c").upsert(data, "enxCPQ__TECH_External_Id__c", { allOrNone: false, allowRecursive: true }, function(err, rets) {
    if (err) {
        reject('error creating attributes: ' + err);
        return;
    }
    let errorsCount = 0;
    for (let i = 0; i < rets.length; i++) {
        if (!rets[i].success) {
            console.log('----- !!! - success: ' + rets[i].success + ' errors: ' + JSON.stringify(rets[i].errors));
            errorsCount++;
        }
    }
    console.log("--- upserted attributes: " + rets.length + ", errors: " + errorsCount);
    resolve();
});
```

borys-kupar:
The same bug with context is in 3 more places: https://github.com/borys-kupar/jsforce/commit/00acfda1bf07569b5ac6270b58634a3b5295dd96

stefangoor:
[...]

PR #985, opened 2020-01-27
Issue #813 fixing missed references to this instead of self

stefangoor:


cristiand391:
Hey, sorry for the late reply, we are just triaging old issues/prs.

Thanks for the PR, unfortunately we aren't doing new feature work on `master` (v1) as we are close to jsforce v3 GA where this was fixed:
https://github.com/jsforce/jsforce/issues/813#issuecomment-2142551227

Issue #883, opened 2019-05-03
No ability to send custom headers with Chatter request


GarethSharpe:
Hi, 

I am currently working with a team to send a Chatter request to Salesforce, however, there is currently no way (to my knowledge) to send headers along with that request. For example, we would like to send { headers: { ""Sforce-Auto-Assign": false } } along with the Chatter resource. Our code currently looks like this:

```
return connection.chatter.resource(AppConstants.FEED_ELEMENT_RESOURCE).create({
        body: { messageSegments },
        feedElementType: AppConstants.FEED_ITEM,
        subjectId: ids[index],
      });
```

The source code of the resource request appears to have a `params.headers` property:

```
/**
 * Sending request to API endpoint
 * @private
 */
Chatter.prototype._request = function(params, callback) {
  if (/^(put|post|patch)$/i.test(params.method)) {
    if (_.isObject(params.body)) {
      params.headers = {
        "Content-Type": "application/json"
      };
      params.body = JSON.stringify(params.body);
    }
  }
  params.url = this._normalizeUrl(params.url);
  return this._conn.request(params, callback);
};
```

Any way to make `params.headers` available to us as well? 

Cheers

PR #996, opened 2020-03-13
Add support for custom headers in Chatter request #883

shunkosa:
Added support for adding custom headers in Chatter requests to send headers like `{ 'X-Connect-Theme': 'Salesforce1' }` to retrieve themed default icons.

Referenced issue: #883


cristiand391:
hey, sorry for the late reply, triaging old issues/prs...

we are near jsforce v3 GA and haven't done much feature work on v1, I'm going to close this in favor of https://github.com/jsforce/jsforce/discussions/1443
rather than exposing an option for each method we should define a way to pass headers (probably in the connection).

PR #998, opened 2020-03-17
Conditionally use PATCH or POST http method for upsert

### LINKED ISSUE #667: Connection.upsert for Inserting New Records Using Id as the External ID
I encountered an error `METHOD_NOT_ALLOWED: HTTP Method 'PATCH' not allowed. Allowed are GET,HEAD,POST` when `connection.sobject('Account').upsert({Name: 'foo'}, 'Id')`.

I found description below in developer guide.
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm

> Inserting New Records Using Id as the External ID
> This example uses the POST method as a special case to insert a record where the Id field is treated as the external ID. Because the value of Id is null, it’s omitted from the request. This pattern is useful when you’re writing code to upsert multiple records by different external IDs and you don’t want to request a separate resource. POST using Id is available in API version 37.0 and later.

So if external ID field is Id and Id is not present, HTTP method for upsert should be `POST`.

These lines:
https://github.com/jsforce/jsforce/blob/a99a38b55346388f53a47d8b2e5afcace43c232b/lib/connection.js#L754

```javascript
      var url = [ self._baseUrl(), "sobjects", sobjectType, extIdField, extId ].join('/');
      return self.request({
        method : 'PATCH',
        url : url,
```

should be like:

```javascript
      var url = _.compact([ self._baseUrl(), "sobjects", sobjectType, extIdField, extId ]).join('/');
      return self.request({
        method : (extIdField.toUpperCase() === 'ID' && extId == null) ? 'POST' : 'PATCH',
        url : url,
```

douglascayers:
Fixes #667 by using `PATCH` or `POST` http method based on the existence of an external id value for the endpoint.

This allows developers, [per these docs](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm), to perform an upsert and specify either the standard `Id` field or an external id to upsert by, don't provide an external id value (to always cause an insert) but **take advantage of** upsert's ability to set lookup fields based on those object's external ids.

### Example 1: No External Id Value

```js
const fooRecord = {
  Bar__r: {
    BarExtId__c: 'baz'
  }
};
conn.upsert('Foo__c', record, 'Id');
```

In the first example, because no `Id` property is on the `fooRecord` then connection.js won't see a value to include in the endpoint URL and then the [Upsert API](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_upsert.htm) endpoint will require POST as the http method.

### Example 2: Yes External Id Value

```js
const fooRecord = {
  FooExtId__c: '42',
  Bar__r: {
    BarExtId__c: 'baz'
  }
};
conn.upsert('Foo__c', record, 'FooExtId__c');
```

In the second example, because `FooExtId__c` property is on the `fooRecord` then connection.js will see a value to include in the endpoint URL and then the [Upsert API](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_upsert.htm) endpoint will require PATCH as the http method.


redwarewolf:
We need this, any updates?

gitmatheus:
Hi team, any updates on this issue?
Thanks.

douglascayers:
@mshanemc Is this something that could get merged and pushed out in a minor release for the `1.x.x` package? 🙏

PR #1012, opened 2020-05-11
Update connection.js

lekkimworld:
Change default API version to latest GA version currently (Spring 20). It's not obvious when using the library that the default API version is from Spring 18 (v42.0) limiting your from using some of the newer capabilities.


cristiand391:

for ref, jsforce v3 defaults to v50.0

PR #1015, opened 2020-05-18
Update dependencies to latest available versions as of May 2020

dhagberg-sf:
- Update jsforce version to 1.10.0

Dependencies:
- base64-url: 2.2.0 => 2.3.3
- coffeescript: 1.10.0 => 2.5.1
- commander: 2.9.0 => 5.1.0
- csv-parse: 4.6.3 => 4.9.0
- csv-stringify: 1.0.4 => 5.5.0
- faye: 1.2.0 => 1.2.5
- inherits: 2.0.1 => 2.0.4
- lodash: 4.7.15 => 4.7.15
- multistream: 2.0.5 => 4.0.0
- opn 5.3.0 => open: 7.0.3 (note: package change)
- promise: 7.1.1 => 8.1.0
- readable-stream: 2.1.0 => 3.6.0
- request: 2.72.0 => 2.88.2
- xml2js: 0.4.16 => 0.4.23

DevDependencies:
- async: 1.5.2 => 3.2.0
- browserify: 16.2.2 => 16.5.1
- bufferutil: 4.0.0 => 4.0.1
- envify: 3.4.1 => 4.1.0
- gulp: 4.0.0 => 4.0.2
- gulp-rename: 1.4.0 => 2.0.0
- gulp-sourcemaps: 1.12.1 => 2.6.5
- gulp-uglify 1.5.4 => gulp-terser 1.2.0 (Note: minifier/uglifier change)
- jsdoc: 3.6.3 => 3.6.4
- mocha: 5.2.0 => 7.1.2
- power-assert: 1.6.0 => 1.6.1
- puppeteer: 1.18.1 => 3.0.4
- testem: 2.8.2 => 3.1.0
- through2: 2.0.3 => 3.0.1
- utf-8-validate: 5.0.1 => 5.0.2

Passes same tests as jsforce master branch (fails same tests also)


package.json
```diff
@@ -32,7 +32,9 @@
     "test:all": "npm run test:node && npm run test:browser",
     "test:node": "mocha --retries 2 --reporter spec --require intelli-espower-loader --exit",
     "test:browser": "testem",
-    "prepublish": "node -e \"if(process.env.npm_package_version!==require('./lib/VERSION')){console.error('The pacakge.json version is not matching to ./lib/VERSION.js');process.exit(1)}\""
+    "prepublish": "node -e \"if(process.env.npm_package_version!==require('./lib/VERSION')){console.error('The pacakge.json version is not matching to ./lib/VERSION.js');process.exit(1)}\"",
+    "snyk-protect": "snyk protect",
+    "prepare": "npm run snyk-protect"
```
amphro, line 37: Will this fail a publish if `snyk protect` finds anything? I'm not sure that is a good thing as some vulnerabilities are addressed at different times.

.snyk
```diff
@@ -0,0 +1,8 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.14.1
+ignore: {}
+# patches apply the minimum changes required to fix a vulnerability
+patch:
+  SNYK-JS-LODASH-567746:
+    - lodash:
+        patched: '2020-05-11T20:10:49.155Z'
```
amphro, line 8: I don't really love this sort of monkey patching. How sever is the vulnerability and is there an issue on lodah that has traction to fix it?

dhagberg-sf: Yeah, not a big fan either.  Here's what I know:

- Medium severity but *can* be attacked over the network (JSON deserialization potentially) : "The function `zipObjectDeep` can be tricked into adding/modifying properties of the Object prototype

- zipObjectDeep is only referenced in the documentation of jsforce.js and jsforce-core.js but is not directly used in any jsforce calls.

- PR: https://github.com/lodash/lodash/pull/4759 (much begging for a release) and alternative PR https://github.com/lodash/lodash/pull/4745

I'm at a bit of a loss regarding the right way forward given the lack of a patched lodash release.

However, I haven't been able to put together an experiment to determine what happens in /downstream/ projects that now pull in unpatched lodash as a transitive dependency - I think they'll need to have the same ignore/suppress setting in their .snyk file.

package.json
dhagberg-sf: What's interesting with the way snyk sets this up, an `npm install` in a /downstream/ project that includes `jsforce` as a dependency will execute this `prepare` statement and monkey-patch lodash (from the .snyk policy file) in the downstream project's npm_modules/ dir.

From what I can tell, the `protect` phase does not check for new vulnerabilities or error out if new vulnerabilities are found.

References:
- https://docs.npmjs.com/misc/scripts#prepublish-and-prepare
- https://support.snyk.io/hc/en-us/articles/360000920818

.snyk
mikelax: lodash finally merged the above PR and released a new version so this specific monkey patch for lodash can be removed.

package.json
dhagberg-sf: Removed snyk bits since lodash update .19 is now published.

.snyk
dhagberg-sf: Removed .snyk bits since `lodash` updated and monkey patching no longer needed.

PR #1063, opened 2020-09-14
feat(conn): batch describe, if-modified-since

jgellin-sf:
This adds support for describing (and caching) multiple SObject types in a single call. It also adds support for the If-Modified-Since header when describing an SObject, which should improve performance when consuming code wants to update an SObject but only if it has been updated since some timestamp.

`Connection.prototype.batchDescribe` and synonym `Connection.prototype.batchDescribeSObjects` are added. This function takes an array of types and a callback as arguments. It posts a `composite/batch` request containing GET requests for each type in the arguments. It returns an array of `DescribeSObjectResult`.

`Connection.prototype.describe` is modified to accept a type (as before) or an options object as a parameter. The options object contains two properties: `type` - the name of the type - and `ifModifiedSince` - a datestamp which is passed to the request as the 'If-Modified-Since' header. If the SObject has not been modified recently enough, the function resolves to undefined.

`Cache.prototype.makeResponseCacheable` is modified to support cases where an array of keys are individually cached to map to a corresponding array of responses.


lib/connection.js
```diff
@@ -1137,28 +1151,112 @@ Connection.prototype.search = function(sosl, callback) {
...
+ * Result returned by batchDescribeSObjects call
+ *
+ * @typedef {Object[]} DescribeSObjectResult
+ */
+/**
+ * Synonym of Connection#batchDescribe()
+ *
+ * @method Connection#batchDescribeSObjects
+ * @param {String[]} types - SObject Type
+ * @param {number} next - the next index in the types array to process (max batch size is 25)
```
amphro, line 1205: So the consumer has to set the next each time they call this if `types` is more than 25?

test/connection-meta.test.js
```diff
@@ -49,6 +60,57 @@ describe("connection-meta", function() {
...
+    before(function (done) {
+      this.timeout(600000); // set timeout to 10 min.
+      testEnv.establishConnection(conn, done);
+    });
+
+    /**
+     *
+     */
+    describe("describe Account, Contact", function () {
+      it("should return metadata information", function (done) {
```
amphro, line 83: Is it possible to test the greater than 25 case?

lib/connection.js
jgellin-sf: Yes. I was back and forth about this, and I'm still on the fence. This method could handle the multiple batch requests itself, but it would make it harder to debug when a problem occurs.

amphro: You could maybe have it be an option, like autofetch is? https://github.com/jsforce/jsforce/blob/master/lib/query.js#L29

test/connection-meta.test.js
jgellin-sf: Yes. See the `test(conn): add test for >25 types` commit.

lib/connection.js
jgellin-sf: Ok I tried it that way. See the `feat(conn): autofetch for batchDescribe` commit.

lib/connection.js
```diff
@@ -1137,28 +1160,145 @@ Connection.prototype.search = function(sosl, callback) {
...
+              );
+            }
+          } else {
+            sobjects.push(subResp.result);
+          }
+          i++;
+        }
+      }
+      if (typesRemaining.length > 0 && autofetch) {
+        return this.doBatchDescribeRequests({ types: typesRemaining, autofetch }).then(
```
xyc, line 1293: We could still look into taking advantage of parallelization here.

xyc: Ref: W-8035201

amphro (approved):
Nice work!

lib/connection.js
```diff
@@ -159,13 +163,32 @@ var Connection = module.exports = function(options) {
...
+      var typesToFetch = types.length > MAX_BATCH_REQUESTS 
+        ? (autofetch ? types : types.slice(0, MAX_BATCH_REQUESTS))
+        : types;
+      var keys = [];
+      typesToFetch.forEach(function (type) { keys.push('describe.' + type); });
+      return keys;
+    }
+  };
+  this.batchDescribe = this.cache.makeResponseCacheable(this.batchDescribe, this, batchCacheOptions);
+  // this.batchDescribeSObjects = this.batchDescribe;
```
amphro, line 190: Why is this commented out? Should it be removed?

jgellin-sf: User error. I uncommented it in `fix(conn): accidentally commented out line of code`.

jgellin-sf: NIce catch @amphro

lib/connection.js
jgellin-sf: Parallelized with `feat(conn): parallelize batch request` commit.

jgellin-sf:
@xyc and @amphro  - I think I've added the last of my changes with that parallelization bit. Will you please give this one last look, and when ci finishes up we can merge?

lib/connection.js
```diff
@@ -1137,28 +1160,151 @@ Connection.prototype.search = function(sosl, callback) {
...
+  var self = this;
+  var types = options.types;
+  var autofetch = options.autofetch || false;
+  var batches = [];
+  do {
+    var batch = types.length > MAX_BATCH_REQUESTS ? types.slice(0, MAX_BATCH_REQUESTS) : types;
+    batches.push(batch);
+    types = types.length > MAX_BATCH_REQUESTS ? types.slice(MAX_BATCH_REQUESTS) : [];
+  } while (types.length > 0 && autofetch);
+  Promise.all(
```
amphro, line 1257: We still want to return this, don't we?

Also, ca we run into max concurrent [API request limits](https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_api.htm)? I guess it would have to be over 625 objects? That doesn't seem that crazy to hit that limit.

jgellin-sf: We do want to return it, yes. I made the change in `fix(conn): return the promise`. Certainly there are orgs with over 625 objects. Should I limit the concurrent requests to batches of 25 (or fewer)?

lms15:
How do I stop this ? Delete this I cannot do it

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

lib/connection.js
amphro: Probably. Can we make it configurable? Maybe in batches of 15 by default. I'm choosing 15 because there could be other tools and services making API calls. I'm suggesting configurable to account for DE and trail orgs. It would probably be pretty rare for a trial org to have over 625 objects but I bet it would happen eventually.

jgellin-sf: Ok I gave it a try. The option `maxConcurrentRequests` has a default and max of 15. The logic is getting pretty complex now. The function sets up the request batches (each length 15 or less) of batch requests(each length 25 or less). Then sequentially (via recursion in promise chain) sets off a batches of concurrent (via Promise.all()) requests.

xyc: @jgellin-sf 🎉 Awesome that you are taking care of parallelization in this PR.

@amphro Worth noting that max concurrent API request limits only applies to long running requests:
> concurrent requests (calls) **with a duration of 20 seconds or longer**.

So I suppose it would be a rare case and if we handle the error properly it should be okay. I think the actual fail-safe way to handle it is to wrap each parallelized promise with a retry block so when it's blocked by the limit it's guaranteed to retry and fetch the results when the other long running requests finish, but it might not be necessary.

lib/connection.js
```diff
@@ -1248,23 +1252,60 @@ Connection.prototype.batchDescribe = Connection.prototype.batchDescribeSObjects
...
+Connection.prototype.doBatchDescribeRequestBatches = function(requestBatches) {
+  // make each batch of requests sequentially
+  var self = this;
+  var sobjects = [];
+  var firstBatch = requestBatches.shift();
+  return self.doBatchOfBatchDescribeRequests(firstBatch).then(
+    function (sobjectArray) {
+      sobjectArray.forEach(function (sobject) { sobjects.push(sobject); });
+      if (requestBatches.length > 0) {
+        return self.doBatchDescribeRequestBatches(requestBatches).then(
```
amphro, line 1281: This is kinda an interesting way to do the recursion. Couldn't you simplify this by just passing in the sobjects and condensing the functions?

```
Connection.prototype.doBatchDescribeRequestBatches = function(requestBatches, sobjects) {
  // make each batch of requests sequentially
  var self = this;
  var requestBatch = requestBatches.shift();

  if (!sobjects) sobjects = [];

  if (requestBatch) {
    return Promise.all(
      requestBatch.map(function (batch) { return self.doBatchDescribeRequest(batch); } )
    ).then(function (results) {
      results.forEach(function (sobjectArray) {
        sobjectArray.forEach(function (sobject) { sobjects.push(sobject); })
      });
      .then(function() {
        return this.doBatchDescribeRequestBatches(requestBatches, sobjects);
      });
  } else {
    return Promise.resolve(sobjects)
  }
}
```

Although, I haven't tested that and "simplified" in this case may be subjective.

amphro (approved):

lib/connection.js
amphro: That's interesting. I didn't even notice that. I thought I ran into this before with short-lived calls before but I just ran some test and sure enough. 

I think having each one in a retry logic is a little overkill, but given the way concurrent request work maybe this is overkill too. @jgellin-sf, I know you already did the work =/ but do you still think it is worth it?

PR #1070, opened 2020-09-30
Fix update-, create- and destoryParallel catch handlers throw null reference error

Codeneos:
This PR addresses an issues in the catch-handlers of `updateParallel`, `createParallel` and `destoryParallel` methods in `connection.js`. The catch handlers call `this._toRecordResult` but as this will not be set for anonymous functions unless they are bound this causes a null reference error. 

This can occurs when two jsforce clients ate deleting a large number or records in parallel causing an error in Salesforce which is the propagated back to the jsforce which then enters the catch handler of `destoryParallel` generating a null-ref error as `this._toRecordResult` to generate an undefined error as `this` is undefined in the context of the catch-handler.

Instead the catch-handler should call `self._toRecordResult`. The problem is repeated for updateParallel`, `createParallel` and `destoryParallel`.

This change simply changes `this._toRecordResult`  -> `self._toRecordResult`; there should be no adverse impact of these changes.


lms15:
I tried to delete the account

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Codeneos:
@stomita Even though work is done on 2.0 is it possible to merge this PR into master? Right now I have this patched using patch-package but it is not ideal. The this reference in the original code is clearly a typo due to which you cannot handle exceptions for `updateParallel`, `createParallel` and `destoryParallel` :(

cristiand391:
hey, sorry this didn't got any response in time. We are close to v3 GA and there's no feature work on v1 so I'm going to close this.

PR #1125, opened 2021-03-01
Add PK Chunking support (Sforce-Enable-PKChunking)

Codeneos:
This PR adds support for PK Chunking (sending the `Sforce-Enable-PKChunking` header) for bulk queries. Three years ago there was a similar PR which right now seems to much out of sync with the master branch to merge. _Note this PR was not based on the previous PR but created from scratch._

The main changes are in the `load` on the `bulk` prototype. It works by allowing to pass an optional options-object as parameter to the `query` function. The two additional options that can be passed:
1. `options.pkChunking` Sets the `Sforce-Enable-PKChunking` to true when `options.chunkSize` is not passed.
2. `options.chunkSize` Sets the `Sforce-Enable-PKChunking`  with a `chunkSize`.

If you set `options.chunkSize` `options.pkChunking` is automatically set to true; they are separate options to allow for using pkChunking with the default chunk size. 

In the `load` function we detect based on the job options if we are dealing with a chunked or a normal batch. If pkCunking is active we wait for the initial batch to get updated to state `NotProcessed`. Once the initial batch is `NotProcessed` we will retrieve the chunked batches (using `job.list`) and poll each batch (chunk) separately. Once _all_ chunks in the job are completed the initial batch emits a response event which then ensures the job is cleaned up and the results are returned. 

The chunks can subsequently be retrieved and made lazy using the lazy-loading feature of the mulitstream package. This prevents the streams from closing before they are even read.

### DIFF
--- a/lib/api/bulk.js
+++ b/lib/api/bulk.js
@@ -805,30 +805,41 @@ Bulk.prototype.load = function(type, operation, options, input, callback) {
     }
   };  
   if (isChunking) {
-    var pollBatches = function(batchInfos) {
-      var batches = batchInfos
-        .filter(function(info) { return info.id != batch.id; })
-        .map(function(info) {
-          var batch = job.batch(info.id);
-          batch.poll(job.pollInterval, job.pollTimeout);
-          return batch.execute();
-        });
-      Promise.all(batches).then(function(res) {
+    var retrieveAll = function(batches) {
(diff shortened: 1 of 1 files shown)


cristiand391:

we suggest to use bulk v2 as it has handles pkchunking automatically, see: https://github.com/jsforce/jsforce/discussions/1433#discussioncomment-10063833

Issue #665, opened 2017-07-13
Support 


tnoonan-salesforce:
@stomita Any reason why the jsforce oauth2 web server auth doesn't support code_verifier and code_challenge? Just resource and priority?

If you ask me that feature should be required on the Salesforce side. It's the only solid way to really protect against replay attacks.

https://help.salesforce.com/articleView?id=remoteaccess_oauth_web_server_flow.htm&language=en&type=0

https://releasenotes.docs.salesforce.com/en-us/spring15/release-notes/rn_forcecom_security_nonce_oauth.htm

It would be nice to also offer an optional auto challenge configuration in oauth.js. If specified the authorization URL generation and the requestToken method will handle the codeVerifier state and encoding automatically for you.

To work around this issue developers will have to override _postParams in oauth.js.

PR #1142, opened 2021-05-28
fix: add more types and exports

amphro:
This PR updates types to cause less breaking changes with existing jsforce projects that use `@types/jsforce`. Specifically, it exports more types and also updates the objects not as the default export.

For example:
```typescript
import Connection from 'jsforce/lib/connection';
```

Can now import using:
```typescript
import { Connection } from jsforce/lib/connection';
```

This has two big benefits. First, all types can be exported, including the default, using `export * from './connection` which is especially useful in `core.js`. Second, it allows renaming in the case of conflicts using `import { Connection as JsforceConnection} from 'jsforce/lib/connection`;

Additionally, this PR adds and changes some parameterized types. The biggest impact one is `R = any` => `R = unknown` on the request methods. Using any should be avoided, and using unknown forces the consumer to think about the type they are requesting. If someone really needs any, simply do `request<any>(...)`.

Lastly, this PR also implements #665. I could break this out into another PR but it made working with the types a little easier.

Bonus: a little bit of code to be able to run some of the test with scratch orgs. Specifically, adding `SF_LOGIN_URL` to the connection manager.


src/oauth2.ts
```diff
@@ -94,12 +110,27 @@ export default class OAuth2 {
     } else {
       this._transport = new Transport();
     }
+    if (useVerifier === true) {
```
mdonnalley, line 113: any reason why you need to explicitly checks that it equals `true`? as opposed to just `if (useVerifier)`

src/process.ts
```diff
@@ -117,7 +118,8 @@ export class ApprovalProcess<S extends Schema> {
    * Get all approval process definitions registered to sobjects
    */
   async list() {
-    const res = await this._conn.request('/process/approvals');
+    // TODO fix type
+    const res = await this._conn.request<any>('/process/approvals');
```
mdonnalley, line 122: I see several of these `TODO fix type` comments. Is there any reason we can't fix them now?

amphro: I have no idea what these APIs return. I can go digging.

mdonnalley (approved):

PR #1152, opened 2021-06-17
Fix unhandled request promise rejections

dzuelke:
When making a request that fails e.g. because of an invalid `instanceUrl`, the resulting error is otherwise not catchable by the caller.

Minimum reproduce case (with nothing on localhost:8000):

```JavaScript
const jsforce =  require("jsforce");
conn = new jsforce.Connection({
  accessToken: "",
  instanceUrl: "http://localhost:8000",
  version: "51.0",
  callOptions: {
    client: "foobar"
  }
});
const response = conn.query("SELECT foo from bar");
response.then(() => console.log("yay")).catch((err) => console.log("an error occurred that we caught:", err));
```

It's catchable with the change, but not without:

```
…/node_modules/node-fetch/lib/index.js:1461
      reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
             ^
FetchError: request to http://localhost:8000/services/data/v51.0/query?q=SELECT%20foo%20from%20bar failed, reason: connect ECONNREFUSED 127.0.0.1:8000
    at ClientRequest.<anonymous> (…/node_modules/node-fetch/lib/index.js:1461:11)
    at ClientRequest.emit (node:events:369:20)
    at Socket.socketErrorListener (node:_http_client:462:9)
    at Socket.emit (node:events:369:20)
    at emitErrorNT (node:internal/streams/destroy:188:8)
    at emitErrorCloseNT (node:internal/streams/destroy:153:3)
    at processTicksAndRejections (node:internal/process/task_queues:81:21) {
  type: 'system',
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED'
}
```

This is for `localhost:12345`, where nothing is listening; when it's e.g. `jdskfhdskjfhasjkdhaksgdkjsahdjksa.com` instead, you'd get an equally uncatchable "`reason: getaddrinfo ENOTFOUND jdskfhdskjfhasjkdhaksgdkjsahdjksa.com`".


dzuelke:
Thanks for the merge @stomita!

I was wondering, does `src/browser/request.ts` need the same change?

And do you think it's worth trying to add a test case for this?

PR #1176, opened 2021-08-24
Expose the HTTP body on the Error object in case there are multiple errors

KeKs0r:
Validation Errors usually have multiple errors as an array. Currently JSforce only exposes the first one.
By attaching the body to the error, it is possible to at least handle multiple errors in userland.


cristiand391:

We've recently fixed this in v3 to throw a specific error code:
https://github.com/jsforce/jsforce/pull/1522

PR #1204, opened 2021-12-17
Merge v1.11.0 from Jsforce master

syzhang88:


syzhang88:
wrong direction wrong

PR #1241, opened 2022-06-10
Update the Faye typing to include other common client methods 

gbockus-sf:
Ran into the issue in a couple of projects where there are missing methods on the current Faye module.  Just extending the existing type to include those. 

I Ran the build locally and attempted to test against a fresh dev org, but had 11 failed tests.
```
Test Suites: 3 failed, 16 passed, 19 total
Tests:       11 failed, 6 skipped, 186 passed, 203 total
Snapshots:   0 total
Time:        387.473 s
```

Did I miss a setup step maybe?


gbockus-sf:
Oh also I had an update to the `package-lock.json` file that I didn't include in the commit.  I assume it was caused by me having a different version of npm. Mostly removing dev properties and integrity value changes.  Let me know if that should be committed.

mdonnalley (approved):

PR #1249, opened 2022-06-30
feat: migrate JWT auth class to jsforce

WillieRuemmele:
Migrate the `JwtOAuth2` class from `sfdx-core` to `jsforce` so that other `jsforce` consumers can use that too
migrate `restDeploy` and `quickDeploy` methods too


src/jwtOAuth2.ts
```diff
@@ -0,0 +1,25 @@
+import OAuth2, { OAuth2Config } from './oauth2';
+
+export type JwtOAuth2Config = OAuth2Config & {
+  privateKey?: string;
+  privateKeyFile?: string;
+  authCode?: string;
+  refreshToken?: string;
+  loginUrl?: string;
+  username?: string;
+};//here
```
paustint, line 10: Is this `//here` comment supposed to be included?

paustint (changes requested):

WillieRuemmele:
https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/client/metadataApiDeploy.ts#L317
https://github.com/salesforcecli/plugin-source/blob/main/src/deployCommand.ts#L181

can be simplified once this is merged

src/api/metadata.ts
```diff
@@ -337,6 +337,43 @@ export class MetadataApi<S extends Schema> {
...
+   *
+   * @param options.id = the deploy ID that's been validated already from a previous checkOnly deploy request
+   * @param options.rest = a boolean whether or not to use the REST API
+   */
+  public async deployRecentValidation(options: {
+    id: string;
+    rest?: boolean;
+  }): Promise<string> {
+    const rest = options.rest;
+    delete options.rest;
```
shetzel, line 351: Doesn't look like you need to delete `options.rest`.  You can probably just do:
`const { id, rest } = options;`
and use them below.

src/jwtOAuth2.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  public jwtAuthorize(innerToken: string): Promise<any> {
+    return super._postParams({
+      grant_type: 'urn:ietf:params:oauth:grant-type:jwt-bearer',
+      assertion: innerToken,
+    });
+  }
+}
+
+export default JwtOAuth2;
```
shetzel, line 25: nit: you can move this to the class definition to avoid this line.  I.e., line 12 above becomes:
`export default class JwtOAuth2 extends OAuth2 {`

shetzel (approved):
Just some nits (and the package-lock.json changes)

PR #1251, opened 2022-07-07
Bulk API V2

### LINKED ISSUE #707: Support for Bulk API 2.0
From the latest salesforce release, Bulk API 2.0 is newly introduced where OAuth 2.0 is supported. Currently, jsforce supporting bulk api 1.0. 

Is there any plan to support bulk api 2.0 in near future.

However, with existing jsforce implementation I am able to use bulk apis using oAuth 2 access token? Is this supported approach or is this working unintended?

colincasey:
Bulk V2 Implementation

Includes:
* support for ingest jobs
* support for query jobs
* hybrid-mode for bulk update and delete operations

closes #707, #1251

### DIFF
--- a/src/api/bulk.ts
+++ b/src/api/bulk.ts
@@ -8,6 +8,7 @@ import joinStreams from 'multistream';
 import Connection from '../connection';
 import { Serializable, Parsable } from '../record-stream';
 import HttpApi from '../http-api';
+import { StreamPromise } from '../util/promise';
 import { registerModule } from '../jsforce';
 import { Logger } from '../util/logger';
 import { concatStreamsAsDuplex } from '../util/stream';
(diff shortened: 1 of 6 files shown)


src/http-api.ts
```diff
@@ -189,7 +189,7 @@ export class HttpApi<S extends Schema> extends EventEmitter {
    * @protected
    */
   async getResponseBody(response: HttpResponse) {
-    if (response.statusCode === 204) {
+    if (response.statusCode === 204 || response.statusCode === 201) {
```
RodEsp, line 192: Heya @colincasey, why would we want to return no content for a 201?
201 is not a no-content reponse: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201. According to those docs it should return the newly crated resource in the body of the message.

src/api/bulk.ts
```diff
@@ -39,13 +44,46 @@ export type BulkOptions = {
...
+  apiVersion: string;
+  assignmentRuleId?: string;
+  columnDelimiter:
+    | 'BACKQUOTE'
+    | 'CARET'
+    | 'COMMA'
+    | 'PIPE'
+    | 'SEMICOLON'
+    | 'TAB';
+  concurrencyMode: string;
```
RodEsp, line 70: It seems like the only mode supported right now is "Parallel", should we just put that here so it can be an enum later when more modes are added?

(Similar to `contentType: 'CSV'` below)

src/api/bulk.ts
```diff
@@ -876,6 +1007,657 @@ export class Bulk<S extends Schema> {
...
+      connection: this.#connection,
+      jobInfo: options,
+      pollingOptions: this,
+    });
+  }
+
+  /**
+   * Create, upload, and start bulkload job
+   */
+  async load(
```
RodEsp, line 1054: Why `load` instead of `upload`? Wouldn't that more accurately describe what this method does?

RodEsp (changes requested):

src/http-api.ts
colincasey: good question.  at one point *i was getting a 201 response with an empty body* but i can't remember what was triggering that now.  it may have been at an earlier stage when i hadn't figured out most of the implementation.  removing it now doesn't seem to affect the test suite so i'll revert this.

src/api/bulk.ts
colincasey: `load` was chosen to maintain consistency with the V1 implementation (see https://jsforce.github.io/jsforce/doc/Bulk.html#load).  i'm open to alternative name suggestions 😄

RodEsp (approved):
@mshanemc are you good with this?

I don't love the name `loadAndWaitForResults` but agree with what @colincasey said [here](https://salesforce.quip.com/CcboAPmPbgYb#TZAADAy1B6x).

Issue #626, opened 2017-04-24
HTTPS_PROXY support


tnoonan-salesforce:
@stomita I see support for setting the HTTP_PROXY environment variable and having it honored by jsforce. However, I'm not seeing support for HTTPS_PROXY. 

Are there technical reasons for this not being supported? Is it just time/resources?

tnoonan-salesforce:
@stomita Poke.

stomita:
HTTP_PROXY can be used for SSL, too.

tnoonan-salesforce:
If HTTPS_PROXY points to a different url than HTTP_PROXY would't that be an issue? We understand that the underlying "request" library that JSforce uses allows them both to be set. So why does JSForce even need to do anything with those env variables?

https://github.com/request/request#proxies

stomita:
As the SF endpoint is always being SSL, why u need distinguish the http/https ?

cristiand391:
This should be fixed in v3, see: https://github.com/jsforce/jsforce/discussions/1529

PR #1256, opened 2022-07-13
fix: support HTTPS proxy env

mshanemc:
sfdx issue with repro: https://github.com/forcedotcom/cli/issues/1626

closes https://github.com/jsforce/jsforce/issues/626

NOTE: this does not provide support for both HTTPS and HTTP proxies simultaneously.  Code in jsforce is using HTTPS
https://github.com/jsforce/jsforce/blob/90a5423c43da3758bdcefb32b17aaa1dc09828a9/src/request.ts#L38

but grabs the env named `HTTP_PROXY`
https://github.com/jsforce/jsforce/blob/90a5423c43da3758bdcefb32b17aaa1dc09828a9/src/transport.ts#L24

this lets you use either, with a preference for `HTTPS_PROXY`

@[W-11426296](https://gus.lightning.force.com/a07EE000011OdYaYAK)@

### DIFF
--- a/src/transport.ts
+++ b/src/transport.ts
@@ -21,7 +21,7 @@ function normalizeApiHost(apiHost: string) {
 }
 
 setDefaults({
-  httpProxy: process.env.HTTP_PROXY ?? undefined,
+  httpProxy: process.env.HTTPS_PROXY ?? process.env.HTTP_PROXY ?? undefined,
   timeout: process.env.HTTP_TIMEOUT
     ? parseInt(process.env.HTTP_TIMEOUT, 10)
     : undefined,


peternhale (approved):
@mshanemc we should open a work item that will support both https and http env settings.

cristiand391 (approved):

cristiand391:
**QA notes**:

`yarn link`ed into sfdx-core -> plugin-auth, then tested with `sfdx auth:jwt:grant` command:

✅ `HTTP_PROXY='http://127.0.0.1:8080/' <command>` goes through proxy
✅  `HTTPS_PROXY='http://127.0.0.1:8080/'` goes through proxy
✅ setting both env vars at the same time:
Screen Shot 2022-07-21 at 13 06 48

PR #1264, opened 2022-08-08
Re/update package lock

RodEsp:
This updates the `package-lock.json` to `lockfileVersion` from `1` to `2` since we should be using the latest version of `npm`.

More info about what that means [here](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion).


mshanemc:
approved.  Ack that 1 browser-test isn't working with this change and investigating that is a separate WI

PR #1267, opened 2022-09-08
ROCKET-153: Log network errors only on debug

dnechay:
Jira: [ROCKET-153](https://mixmaxhq.atlassian.net/browse/ROCKET-153)

Log from this PR was unstructured and logged on every occurrence, so it lead to confusion when I though from CloudWatch logs that this error is unhandled and interrupts the main process.
image
Moved this to debug level since we throw it anyway and handle on top level.


dnechay:
Missed

PR #1278, opened 2022-10-19
fix(bulk v2): paginate to get all job results

cristiand391:
Updates `QueryJobV2.getResults` method to paginate through all results in a query job.

Fixes: https://github.com/forcedotcom/cli/issues/1759

### DIFF
--- a/src/api/bulk.ts
+++ b/src/api/bulk.ts
@@ -18,6 +18,7 @@ import {
   HttpResponse,
   Record,
   Schema,
+  Optional,
 } from '../types';
 import { isFunction, isObject } from '../util/function';
 
@@ -1100,9 +1101,11 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
   readonly #operation: QueryOperation;
   readonly #query: string;
   readonly #pollingOptions: BulkV2PollingOptions;
-  #queryResults: Record[] | undefined;
+  #queryResults: Record[][] | undefined;
   #error: Error | undefined;
   jobInfo: Partial<JobInfoV2> | undefined;
+  locator: Optional<string>;
+  finished: boolean = false;
 
   constructor(options: CreateQueryJobV2Options<S>) {
     super();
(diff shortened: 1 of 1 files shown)


src/api/bulk.ts
```diff
@@ -1213,25 +1216,46 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
...
 
-      this.#queryResults = results ?? [];
+  private getResultsUrl() {
+    const url = `${this.#connection.instanceUrl}/services/data/v${this.#connection.version}/jobs/query/${getJobIdOrError(this.jobInfo)}/results`
 
-      return this.#queryResults;
-    } catch (err) {
-      this.emit('error', err);
-      throw err;
+    return this.locator ? `${url}?locator=${this.locator}` : url
```
cristiand391, line 1237: API docs:
https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_get_job_results.htm

src/api/bulk.ts
```diff
@@ -1100,9 +1101,11 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
   readonly #operation: QueryOperation;
   readonly #query: string;
   readonly #pollingOptions: BulkV2PollingOptions;
-  #queryResults: Record[] | undefined;
+  #queryResults: Record[][] | undefined;
```
cristiand391, line 1104: This type was wrong, the results from the parsed body (raw csv) is a 2d array of records.

src/api/bulk.ts
```diff
@@ -1213,25 +1216,46 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
...
-      const results = await this.createQueryRequest<Record[] | undefined>({
-        method: 'GET',
-        path: `/${getJobIdOrError(this.jobInfo)}/results`,
-        responseType: 'text/csv',
-      });
+    const httpApi = new HttpApi(this.#connection, options);
+    httpApi.on('response', (response: HttpResponse) => {
+      this.locator = response.headers['sforce-locator']
+    })
+    return httpApi.request<R>(request_);
```
cristiand391, line 1231: `request` method wraps `HttpApi` so that it can access the locator header from the emitted response
https://github.com/jsforce/jsforce/blob/2.0/src/http-api.ts#L113

src/api/bulk.ts
```diff
@@ -1213,25 +1216,46 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
...
+  }
+
+  async getResults(): Promise<Record[][]> {
+    if (this.finished && this.#queryResults) {
+      return this.#queryResults
+    }
+
+    this.#queryResults = []
+
+    while (this.locator !== 'null') {
```
cristiand391, line 1247: if there are no more results. the locator header is `null` (string), not `null` or `undefined`

src/api/bulk.ts
cristiand391: the returned header is `Sforce-Locator` but jsforce converts all headers to lowercase, see: https://github.com/jsforce/jsforce/blob/2.0/src/request.ts#L63

peternhale (approved):

Issue #842, opened 2018-12-02
Tooling API - Can't create objects


MoshikEilon:
When I try to create an object such as 'ApexClass', I get the following error:

TypeError: this._createSingle is not a function
At: at Tooling.Connection.insert.Connection.create

This is the code:
`
conn.login(username, password + securityToken, function(err) {
    if (!err) {
        const apexBody = [
            "public class TestApex {",
            "  public string sayHello() {",
            "    return 'Hello';",
            "  }",
            "}"].join('\n');
        conn.tooling.sobject('ApexClass').create({
            body: apexBody
          }, function(err, res) {
            if (err) { return console.error(err); }
            console.log(res);
          });
    }
});
`

aust1nz:
I had the same issue with JSforce v 1.9.1. I downgraded to 1.8.5 and was able to use the Tooling API as expected.

7ynk3r:
I got a similar error `this._ensureVersion is not a function` and downgrading solved it as well

preveenraj:
Upgrading to 1.9.3 will be a better move, as dated now.

amphro:
@preveenraj are you staying this is fixed in >1.9.3?

superkensington:
Can confirm that this is still an issue in 2.0.0-beta.19. Unable to use the Tooling API functionality for DML operations, throws the error `TypeError: this._ensureVersion is not a function`.

cristiand391:
This should be fixed in v3, see: https://github.com/jsforce/jsforce/discussions/1529

PR #1289, opened 2022-11-23
Tooling API CRUD fix (#842)

superkensington:
Added missing method to Tooling API class from base Connection class and dependent property. Also added some basic CRUD tests for the Tooling API since none existed, modeled after the Connection CRUD tests. Fixes jsforce/jsforce#842


src/api/tooling.ts
```diff
@@ -181,6 +181,7 @@ export type CompletionsResult = {
 const {
   query,
   queryMore,
+    _ensureVersion,
```
kyungkimsf, line 184: nit: indentation seems to be off for the added lines of code in this file

mosel-slalom: @kyungkimsf is it possible to get this merged soon?

superkensington: > nit: indentation seems to be off for the added lines of code in this file

✅ Fixed

Issue #1294, opened 2022-12-03
DNS timeout on logout (getaddrinfo EAI_AGAIN)


paulsmithkc:
When closing the connection to salesforce, with the logout method. We are sometimes seeing the following error:

```
getaddrinfo EAI_AGAIN xxxx.my.salesforce.com
```

This indicates to me that there is not currently a DNS cache being employed to cache the DNS lookup. Adding a simple DNS cache should speed up requests, take some load of the DNS servers, and avoid these errors.

https://stackoverflow.com/a/73581100/8645145

https://www.npmjs.com/package/cacheable-lookup

paulsmithkc:
@jgellin-sf @stomita Bumping because we saw this same error on connection attempts last night.

cristiand391:
triaging old issues...

jsforce v3 uses `node-fetch` v2 which supports this by default:
https://github.com/node-fetch/node-fetch/tree/2.x?tab=readme-ov-file

> If no agent is specified, the default agent provided by Node.js is used. Note that [this changed in Node.js 19](https://github.com/nodejs/node/blob/4267b92604ad78584244488e7f7508a690cb80d0/lib/_http_agent.js#L564) to have keepalive true by default. If you wish to enable keepalive in an earlier version of Node.js, you can override the agent as per the following code sample.

https://nodejs.org/api/http.html#http_new_agent_options

can someone confirm if this still happens on jsforce v3?

PR #1295, opened 2022-12-03
Keep connection alive for 1 minute

paulsmithkc:
Keep connection alive for 1 minute to improve overall performance. Default behavior was to open a new socket for each request.

Related: #1294


paulsmithkc:
Bump

paulsmithkc:
This PR needs further testing, to see if it is viable. As salesforce has a low connection limit.

Issue #1291, opened 2022-11-30
login call succeeds, but the query call is stuck in the 401 loop


qbert-007:
Is there a chance that the login call would succeed but the query will keep returning a 401?
```
const conn = new jsforce.Connection({
  clientId: clientId,
  clientSecret: clientSecret,
  version: "54.0",
  loginUrl: 'https://login.salesforce.com/',
});

conn.login(username, password, function (error, res) {
    if(error){
        console.log('error in login:', error)
    } else{
        conn.query('SELECT Id, Name, OrganizationType FROM Organization', function (err, result){
            if(err){
                console.log('error in query call:', err)
            }else{
                console.log('success')
            }
        })
    }
})
```

We are facing issue in which the login call is successful but the query call fails with a 401, and then another query call is made and a 401 is received; this occurs in a loop. After few hours we receive the below error in callback:
`{ message: 'authentication failure', name: 'invalid_grant', stack: 'invalid_grant: authentication failure\n' + ' at /app/node_modules/jsforce/lib/oauth2.js:198:19\n' + ' at tryCallOne (/app/node_modules/promise/lib/core.js:37:12)\n' } 
`

qbert-007:
Update: We figured out that when `Session Security Level Required at Login` is set to `High Assurance` in the user's profile, the query call gets stuck in a loop and the callback never triggers.

Salesforce throws a `401 Unauthorized` for the above session setting, and there is no handling done for this case in jsforce, and it keeps refreshing tokens here: https://github.com/jsforce/jsforce/blob/master/lib/http-api.js#L90 and never call the callback fn.

qbert-007:
Can this issue be addressed?

qbert-007:
This issue is critically impacting our integration. @stomita Could this please be acknowledged and fixed?

samschurter:
Running into the exact same issue, came to the same conclusion on the same line number. Is this library maintained anymore or is it dead?

dtto86:
Faced similar issue for `describe()` function call when expired credentials were used in `login()` function.

These PR changes https://github.com/jsforce/jsforce/pull/1345 were able to fix the issue related to expired credentials.

Also, tried this PR https://github.com/jsforce/jsforce/pull/1301 changes. They worked as expected when High Assurance session setting was enabled in User's Profile.

@stomita Is it possible to merge these changes in version 1.x.x or they have been addressed in version 2.x.x?

cristiand391:
[...]

PR #1301, opened 2023-01-09
Handle case when session is not valid for use with the REST API

qbert-007:
When `Session Security Level Required at Login` is set to `High Assurance` level in the user's profile. 
And below default setting is set for session security level:
image

Salesforce throws below error when doing query call:
```
{
  statusCode: 401,
  headers: {
    "cache-control": "no-cache,must-revalidate,max-age=0,no-store,private",
    connection: "close",
    "content-type": "application/json;charset=UTF-8",
    date: "Fri, 06 Jan 2023 09:33:41 GMT",
    "set-cookie": "CookieConsentPolicy=0:1; path=/; expires=Sat, 06-Jan-2024 09:33:41 GMT; Max-Age=31536000, LSKey-c$CookieConsentPolicy=0:1; path=/; expires=Sat, 06-Jan-2024 09:33:41 GMT; Max-Age=31536000, BrowserId=NNa2EI2lEe2R9MHZgNtMtQ; domain=.salesforce.com; path=/; expires=Sat, 06-Jan-2024 09:33:41 GMT; Max-Age=31536000",
    "strict-transport-security": "max-age=63072000; includeSubDomains",
    "transfer-encoding": "chunked",
    "x-content-type-options": "nosniff",
    "x-robots-tag": "none",
    "x-xss-protection": "1; mode=block",
  },
  body: "[{\"message\":\"This session is not valid for use with the REST API\",\"errorCode\":\"INVALID_SESSION_ID\"}]",
}
```
To handle this, we need to check whether the session is recoverable before re-sending request with refreshed token. Else it get stuck in circular pattern by sending request again and again with refreshed token.

Related issue: https://github.com/jsforce/jsforce/issues/1291


cristiand391:
we fixed this for SOAP login:
https://github.com/jsforce/jsforce/issues/1411

I can't 100% tell that all responses with `This session is not valid for use with the REST API` aren't valid for refresh, if someone finds a way to consistently repro this please log a new issue (I tried whit OP suggestion and others but never got the same API response to test it).

PR #1306, opened 2023-02-13
fix: expose line ending

peternhale:
@W-12537051@


WillieRuemmele (approved):
seems right to me. Any tests or way to verify it's fixing the bug?

src/api/bulk.ts
```diff
@@ -1330,6 +1330,7 @@ export class IngestJobV2<
           externalIdFieldName: this.jobInfo?.externalIdFieldName,
           object: this.jobInfo?.object,
           operation: this.jobInfo?.operation,
+          lineEnding: this.jobInfo?.lineEnding,
```
cristiand391, line 1333: What happens if the caller doesn't specify a line ending? docs says it defaults to `LF` if it's not specified, does the same happens if the field in the body passed `undefined`?

peternhale: I can default it based on platform

cristiand391: I'm not sure how webpack handles the `process.env` object so not sure what's needed for it to replace it in the browser bundle. Maybe just default the param to `LF`?

peternhale:
> 

@WillieRuemmele it all get passed through to the backend.

WillieRuemmele (approved):

src/api/bulk.ts
cromwellryan: @cristiand391 it's using [envify](https://github.com/jsforce/jsforce/blob/master/gulpfile.babel.js#LL106C8-L106C14)

cristiand391: @cromwellryan that's the `master`/v1 branch, I don't see that in the  `2.0` branch:
https://github.com/jsforce/jsforce/tree/2.0

cristiand391: We tested this by setting `lineEnding = undefined` and the server ignores it and defaults to `LF` as expected.

cristiand391 (approved):

Issue #990, opened 2020-02-12
Bullk job creation URI error


RoshanRivi:
Hi Dev's,

I'm trying to use jsForce for bulk operations. I'm creating a conn using accessToken, instanceUrl and passing it to this function, but it is trowing `Error: Invalid URI "/services/async/42.0/job"`. I have tried other operations, everything returns me same wherever I use `conn.bulk`.

I have checked the API version using postman it is working. Everything works perfectly, I was able create Job and run the batch through postman.

Can you check and confirm whether it is function or repo has this issue

```
export const writeAccounts= async (conn) => {
 var accounts = [
    { Name: "Account #1" },
    { Name: "Account #2" },
    { Name: "Account #3" }
  ];
  // Create job and batch
  var job = conn.bulk.createJob("Account", "insert");
  var batch = job.createBatch();
  // start job
  batch.execute(accounts);
  // listen for events
  batch.on("error", function(batchInfo) {
    // fired when batch request is queued in server.
    console.log("Error, batchInfo:", batchInfo);
  });
  batch.on("queue", function(batchInfo) {
    // fired when batch request is queued in server.
    console.log("queue, batchInfo:", batchInfo);
    batch.poll(1000 /* interval(ms) */, 20000 /* timeout(ms) */); // start polling - Do not poll until the batch has started
  });
  batch.on("response", function(rets) {
    // fired when batch finished and result retrieved
    for (var i = 0; i < rets.length; i++) {
      if (rets[i].success) {
        console.log("#" + (i + 1) + " loaded successfully, id = " + rets[i].id);
      } else {
        console.log(
          "#" +
            (i + 1) +
            " error occurred, message = " +
            rets[i].errors.join(", ")
        );
      }
    }
    // ...
  });
}
```

Console error.

`Error: Invalid URI "/services/async/42.0/job"`

**Thanks in advance.**

RoshanRivi:
Sorry my fault I was calling it out of conn scope

Issue #979, opened 2020-01-07
Lambda - cannot find user home directory to store configuration files


dylandechant:
trying to get a lambda function up and running but am running into an error just importing it into a file (not even using or calling the library)
here is the error
```json
{
    "errorType": "Error",
    "errorMessage": "cannot find user home directory to store configuration files",
    "stack": [
        "Error: cannot find user home directory to store configuration files",
        "    at getDefaultConfigFilePath (/var/task/node_modules/jsforce/lib/registry/file.js:42:11)",
        "    at new FileRegistry (/var/task/node_modules/jsforce/lib/registry/file.js:64:47)",
        "    at Object.<anonymous> (/var/task/node_modules/jsforce/lib/registry/index.js:47:98)",
        "    at Module._compile (internal/modules/cjs/loader.js:956:30)",
        "    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)",
        "    at Module.load (internal/modules/cjs/loader.js:812:32)",
        "    at Function.Module._load (internal/modules/cjs/loader.js:724:14)",
        "    at Module.require (internal/modules/cjs/loader.js:849:19)",
        "    at require (internal/modules/cjs/helpers.js:74:18)",
        "    at Object.<anonymous> (/var/task/node_modules/jsforce/lib/jsforce.js:38:40)"
    ]
}
```

i understand the cli saves to `~/.jsforce/config.json`, but that doesn't seem to be required for non-cli usage. is there anyway to override this check?

dylandechant:
solved:

to fix this add an environment variable to your lambda function that points it to `/tmp`
`HOME=/tmp`

lms15:
OK thanks glad it was solved

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

mayrbenjamin92:
This might not be the solution for all users of this module! The default should actually be something in the "node_modules" folder itself. What if I can't set an ENV variable?

olup:
This is very wrong and require hacking the env variables, especially when having also a local setup. Could it be allowed to pass this path directly in some jsforce initialization params ?

dohomi:
I also bumped into this issue while trying to use Vercel hosting. Any solution out there?

genk1:
I solved by overriding process.env.JSFORCE_CONNECTION_REGISTRY before importing salefoce library

```
[...]

PR #1313, opened 2023-03-21
W-12597177 - Add support for multipart/form-data requests with connec…

andrewslater:
…tion.js

This change introduces new options to `DmlOptions` where individual fields can be marked to be sent as multipart/form-data elements. This is useful for sending requests that are > 52mb (limit when using application/json content-type). See tests for details on usage. Also note it is unreasonable to expect this code to break any existing consumers of this library because all change in behavior is gated on the use of the new `multipartFileFields` option.


test/tooling.test.ts
```diff
@@ -90,6 +91,57 @@ it('should get completions and return completions', async () => {
...
+        filename: 'test.zip',
+      },
+    },
+  };
+  const record = await conn.tooling.create('StaticResource', request, options);
+  assert.ok(record.success);
+});
+
+it('can create static resource using multipart/form-data content type with a base64 encoded string', async () => {
+  const request = createStaticResourceRequest(
```
ahmedrana2022, line 119: question: wondering why we would need to base-64 encode the file, wouldn't the caller just pass the zip file (like in the previous test) ?

src/connection.ts
```diff
@@ -969,13 +970,41 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
...
+          form.append(
+            fieldName,
+            Buffer.from(rec[fieldName], 'base64'),
+            fileDetails,
+          );
+          delete rec[fieldName];
+        },
+      );
+      // Serialize the remaining fields as JSON
+      form.append(type, JSON.stringify(rec), {
```
ganeshnt, line 990: With my limited knowledge on TS... Logic is in this block of execution only when the conditional at #975 is satisfied, which establish that we are in this block to process multi-part requested fields.  So the the loop starting on #979 iterate to process each field that needs multi-part function.  If my assessment of this logic is correct, then i do not understand what is left to process on line #990?

ganeshnt: if there were fields that need JSON processing, shouldn't this branch into the "else" clause below?

test/data/test.zip
ganeshnt, line 1: what do we have in the zip? and Git shows +348 bytes only!

ganeshnt (approved):
with my limited knowledge on TS and this area of code, the review looks okay to me and i do not see any reason for potential regressions to other areas considering this change is way down the stack

src/connection.ts
andrewslater: What remains is any fields that were not referenced in `multipartFileFields`. The key is the `delete rec[fieldName]` on line 986. Consider a hypothetical `rec` for an entity named `MyEntity" is passed in:
```
{
  Name: "MyRecordName",
  Cacheable: True,
  MyFileField: "<base64 encoded string or buffer>"
}
```

with `options.multipartFileFields`:
```
{
  MyFileFIeld: {
    contentType: "image/png",
    fileName: "company-logo.png"
  }
}
```

After the loop starting on line 979 completes, we have a `FormData` object that contains (when serialized):
```
----------------------------248819749774746952254207
Content-Disposition: form-data; name="MyFileField"; filename="company-logo.png" 
Content-Type: image/png
<base64 encoded file content>
```

We still need to add the remaining fields so line 990 serializes the next "part" of our request which results in:
```
----------------------------248819749774746952254207
Content-Disposition: form-data; name="MyEntity"
Content-Type: application/json
 
{"Name": "MyRecordName", "Cacheable": true}
```

Now all parts will be sent, with the appropriate content type.

test/tooling.test.ts
andrewslater: While iterating on this I had some (incorrect) changes to  `request-helper.ts` that resulted in passing a Buffer to fail but passing a string to work, so I wanted to have tests for both scenarios.

andrewslater: It passes a base64 encoded string that represents a zip file. The content type for a multipart "part" is not the same as the "Content-Type" HTTP header. Each "part" has to specify what its content type is. This is the equivalent of the curl command option you referenced in your research:

```
-F 'versionInfo=@package-version-info.zip;type=application/zip'
```

For binary content types (as opposed to, say, the json "part" we send for the non-file fields) you specify a content type for the _"part"_ and base64 encode the contents. In this case the `application/zip` content type for the `Body` part is used to tell the server what type of data will be present after decoding the base64 encoded string.

I tested that the StaticResource is saved and is identical when retrieved in both cases.

ganeshnt (approved):

test/tooling.test.ts
ahmedrana2022: Got it, thanks for clarifying that.

ahmedrana2022 (approved):

cristiand391 (approved):

PR #1326, opened 2023-05-09
W-12557202 - Support arbitrary response decoding in order to handle binary stream responses

andrewslater:
This PR introduces a new HttpRequestOption `decodeResponseAs` which allows the caller to determine how the response stream should be decoded. Before this change responses were always decoded as utf-8 strings. This remains the default behavior but callers can now specify e.g. `decodeResponseAs: 'base64'` to have a binary response stream decoded into a base64 string that can then be manipulated as needed.


cristiand391 (changes requested):
LGTM. Since we are referring to this prop internally as `encoding`, can we use that instead of `decodeResponseAs`?

That's also how `got` does it:
https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#encoding

andrewslater:
> LGTM. Since we are referring to this prop internally as `encoding`, can we use that instead of `decodeResponseAs`?
> 
> That's also how `got` does it: https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#encoding

Updated PR to use `encoding` instead of `decodeResponseAs`

cristiand391 (approved):

PR #1338, opened 2023-06-01
 chore(ci): add manual release workflow

cristiand391:
Add a GHA workflow to do releases from CI via a manual trigger.

It has a few specific values like `pre-release-identifier` that should be deleted once v2 is GA.
The test pipeline still runs on Circle and doesn't have any integration with this, so once you merge a PR you need to manually trigger a release from the actions tab.

**TODO**:
1) add npm and GH tokens

### DIFF
--- a/.github/workflows/manualRelease.yml
+++ b/.github/workflows/manualRelease.yml
@@ -1,0 +1,40 @@
+name: manual release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+      - name: Conventional Changelog Action
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@d360fad3a42feca6462f72c97c165d60a02d4bf2
+        # overriding some of the basic behaviors to just get the changelog
+        with:
+          git-user-name: svc-cli-bot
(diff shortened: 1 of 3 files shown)


.github/workflows/manualRelease.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Conventional Changelog Action
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@d360fad3a42feca6462f72c97c165d60a02d4bf2
+        # overriding some of the basic behaviors to just get the changelog
+        with:
+          git-user-name: svc-cli-bot
+          git-user-email: svc_cli_bot@salesforce.com
+          github-token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+          output-file: false
+          pre-commit: ./scripts/build.js
```
cristiand391, line 22: use a pre-commit hook to include the bundle changes (`dist` folder) in the release commit.
https://github.com/TriPSs/conventional-changelog-action/#pre-commit-hook

https://github.com/TriPSs/conventional-changelog-action/blob/84d56d6ef8cfa17b0b0e8c059c11b663b4ed0323/src/index.js#L180-L193

This adds the overhead of running `npm run build` twice (one here, then again when publishing to npm) but will allow to have a snapshot of releases in GH that include the browser bundles too.

mshanemc: if we eventually put `wireit` on jsforce that could avoid script repetition when there aren't any code changes, and the maybe lets more of these tasks run in parallel where there aren't dependencies between them.

PR #1343, opened 2023-06-07
chore: fix types for bundling

### LINKED ISSUE #1342: [v2] : unable to bundle due to removed types
v2 does not currently play nice with bundlers.

for example rollup will error with something like 
```
[!] RollupError: "Registry" is not exported by "node_modules/@salesforce/apex-node/node_modules/jsforce/module/registry/types.js", imported by "node_modules/@salesforce/apex-node/node_modules/jsforce/module/registry/index.js".
https://rollupjs.org/troubleshooting/#error-name-is-not-exported-by-module
node_modules/@salesforce/apex-node/node_modules/jsforce/module/registry/index.js (1:9)
```

babel removes the type info from the types files since js does not care about them.
This has the effect of having an import with no export which bundlers do not like.
You can see this most clearly in the [analysis.ts](https://github.com/jsforce/jsforce/blob/2.0/src/api/analytics.ts#L24) file which is re exporting types from the types file.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

This also affects package which this as a dep such as:

1. https://github.com/forcedotcom/sfdx-core and 
1. https://github.com/forcedotcom/salesforcedx-apex/ (which has core as dep)

lukecotter:
add type to type only imports and exports

This is to play nice with bundlers.
babel removes the type info from the types files since js does not care about them.
This has the effect of having an import with no export which bundlers do not like.
You can see this most clearly in the analysis.ts file which 
is re exporting types from the types file.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

resolves #1342

### DIFF
--- a/src/api/analytics.ts
+++ b/src/api/analytics.ts
@@ -21,7 +21,7 @@ import {
 import { QueryExplainResult } from '../query';
 
 /*----------------------------------------------------------------------------------*/
-export {
+export type {
   ReportMetadata,
   ReportExecuteResult,
   ReportRetrieveResult,
--- a/src/core.ts
+++ b/src/core.ts
@@ -18,11 +18,11 @@ export {
   VERSION,
   SfDate as Date,
   SfDate,
-  Registry,
   BrowserClient,
   RecordReference,
   RecordStream,
   registry,
   browser,
 };
+export type { Registry };
 export default jsforce;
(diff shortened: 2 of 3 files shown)


cristiand391 (approved):

cristiand391:
Thanks for the fix @lukecotter!
We'll probably do another v2 release sometime next week.

PR #1345, opened 2023-06-13
Fixes #1291

### LINKED ISSUE #1291: login call succeeds, but the query call is stuck in the 401 loop

braj1999:


cristiand391:

I'm working on a PR to address this in the v3 branch.

PR #1349, opened 2023-06-24
Bring back JWT OAuth2 module (as deprecated)

stomita:
As the latest update of jsforce (beta25), which includes removal of JwtOAuth2 #1341 , `sfdx-cli` is failing for the absence of `jsforce_1.JwtOAuth2`. This will be fixed in the `sfdx-core`, but the fix is not yet merged and it requires all sfdx-cli users to update its version because the sfdx-core is depending jsforce in caret versioning (^2.0.0-beta.24), which automatically select the latest beta version.
Considering the impact of this disfunction, I would like to bring back the JwtOAuth2 module, but with deprecated warning.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -62,7 +62,6 @@ import Metadata from './api/metadata';
 import SoapApi from './api/soap';
 import Streaming from './api/streaming';
 import Tooling from './api/tooling';
-import { JwtOAuth2Config, JwtOAuth2 } from './jwtOAuth2';
 import FormData from 'form-data';
 
 /**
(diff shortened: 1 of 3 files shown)


cristiand391:
Hi @stomita 👋🏼 

I see you published a beta release last week, do you know if that also changed the `latest` dist tag? I see it points to latest v2 beta instead of v1:
```
➜  ~ npm show jsforce dist-tags
{
  latest: '2.0.0-beta.27',
  alpha: '2.0.0-alpha.13',
  beta: '2.0.0-beta.25'
}
```

Also, we added a manual release workflow a few weeks ago that can be used to trigger a release.

We'll add a doc for this once all actions are added, but this is how it works:
1) merge change into `2.0` branch, don't modify `version` field in `package.json`, the action will do it at release time
2) go to https://github.com/jsforce/jsforce/actions/workflows/manualRelease.yml, click `Run workflow` -> `Use workflow from`  and select the `2.0` branch -> `Run workflow`.
This will create a [GH release](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) that triggers the `npm-publish` job.
Screenshot 2023-06-26 at 18 42 58

PR #1353, opened 2023-07-07
ci: add GHA test workflows

cristiand391:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+      - run: |
+          echo Installed Chrome version: ${{ steps.setup-chrome.outputs.chrome-version }}
+          ${{ steps.setup-chrome.outputs.chrome-path }} --version
+
+          CHROME_VERSION=$(echo "${{ steps.setup-chrome.outputs.chrome-version }}" cut -f 1 -d ' ' | cut -d '.' -f 1)
+
+          CHROMEDRIVER_VERSION=$(curl --location --fail --retry 10 http://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROME_VERSION)
+
+          echo "chromedriver_version=$CHROMEDRIVER_VERSION" >> "$GITHUB_ENV"
+      - uses: nanasess/setup-chromedriver@v2
```
mshanemc, line 29: for 3PP actions, use @sha instead of @v2 for security reasons.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Lint and Type Check
+        run: |
+          npm run lint
+          npm run build:types
+          npm run typecheck
+
+      - name: Run Node tests
+        run: npm run test:node
+
+  test-node-windows:
```
mshanemc, line 79: why not make `runs-on` part of the matrix?

mshanemc (approved):

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+on:
+  push:
+    branches-ignore: [2.0]
+  workflow_dispatch:
+
+jobs:
+  test-browser:
+    runs-on: ubuntu-latest
+    env:
+      SF_USERNAME: ${{ secrets.SF_USERNAME }}
```
mshanemc, line 11: these tests need another workflow that will
auth to the hub (like the NUTs do)
create a scratch org (json output will have username, instanceUrl)
`user password generate` to get the password

mshanemc: then all the 7 tests can run in parallel with isolated orgs

.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
         arrowParens: 'always',
         singleQuote: true,
         trailingComma: 'all',
+        endOfLine: "auto",
```
cristiand391, line 30: needed to fix lint on windows.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,155 @@
...
+  
+      - name: Run Node tests
+        env:
+          SF_OAUTH2_CLIENT_ID: ${{ secrets.SF_OAUTH2_CLIENT_ID }}
+          SF_OAUTH2_CLIENT_SECRET: ${{ secrets.SF_OAUTH2_CLIENT_SECRET }}
+        run: |
+          SF_USERNAME=$(sf org display user --target-org jsforce-test-org --json | jq -r '.result.username') \
+          SF_PASSWORD=$(sf org display user --target-org jsforce-test-org --json | jq -r '.result.password') \
+          SF_LOGIN_URL=$(sf org display user --target-org jsforce-test-org --json | jq -r '.result.instanceUrl') \
+          npm run test:node
```
cristiand391, line 104: inline env vars with scratch org credentials (see the backslash at the EOL) to avoid exposing them in workflow logs.
We could also do `echo "SF_USERNAME=$(sf org display user --target-org jsforce-test-org --json | jq -r '.result.username')" >> $GITHUB_ENV` but since they aren't repo secrets they will not be redacted in this section: 
Screenshot 2023-08-01 at 11 09 33

.github/workflows/tests.yml
```diff
@@ -0,0 +1,155 @@
...
+  
+      - name: Delete scratch org
+        if: always()
+        run: sf org delete scratch --target-org jsforce-test-org --no-prompt
```
cristiand391, line 108: I ended up not using an action since we just run the script to setup a scratch org and also because composite actions don't support post steps:
https://github.com/actions/runner/issues/1478

this step will ensure the scratch org always gets deleted even if tests fail.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,155 @@
...
+
+ #  test-node-windows:
+ #    runs-on: windows-latest
```
cristiand391, line 111: leave disabled for now, need to figure out zx/shell not escaping some chars in file paths.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,155 @@
...
+  #       with:
+  #         chrome-version: stable
+  #       id: setup-chrome
+  #     - run: |
+  #         echo Installed Chrome version: ${{ steps.setup-chrome.outputs.chrome-version }}
+  #         ${{ steps.setup-chrome.outputs.chrome-path }} --version
+  #
+  #         CHROME_VERSION=$(echo "${{ steps.setup-chrome.outputs.chrome-version }}" cut -f 1 -d ' ' | cut -d '.' -f 1)
+  #
+  #         CHROMEDRIVER_VERSION=$(curl --location --fail --retry 10 http://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROME_VERSION)
```
cristiand391, line 22: this was working a few weeks ago when 114 was in the stable channel:
https://github.com/jsforce/jsforce/actions/runs/5488741617/job/14865068976

current stable is 115 and the URL to get the chromedriver changed:
https://chromedriver.chromium.org/downloads

Will log an issue in the setup-chromedriver action repo.

cristiand391: also tried pinning to 114 in setup-chrome (then setup-chromedriver will try to get the driver for 114) but was unable to find the commit reference to chrome 144 (action doesn't accept a version).

scripts/org-setup.mjs
```diff
@@ -0,0 +1,84 @@
+import 'zx/globals'
+import { writeFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import { EOL } from 'node:os'
+
+if (!process.env.SF_HUB_USERNAME || !process.env.SF_OAUTH2_JWT_KEY || !process.env.SF_OAUTH2_CLIENT_ID) {
+  throw new Error('Set SF_HUB_USERNAME, SF_OAUTH2_JWT_KEY and SF_OAUTH2_CLIENT_ID env vars.')
+}
+
+$.verbose = false
```
cristiand391, line 10: disable zx printing output of executed commands:
https://github.com/google/zx#verbose

scripts/org-setup.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+const hubOpts = {
+  username: process.env.SF_HUB_USERNAME,
+  clientId: process.env.SF_OAUTH2_CLIENT_ID,
+  jwtKeyFile: join(process.cwd(), 'jwtKey.txt')
+}
+
+await writeFile(hubOpts.jwtKeyFile, formatJwtKey())
+
+await $`sf org login jwt --username ${hubOpts.username} --jwt-key-file "${hubOpts.jwtKeyFile}" --set-default-dev-hub --client-id ${hubOpts.clientId}`
```
cristiand391, line 20: zx will throw if a command returns a non-zero exit code:
https://github.com/google/zx#command-

scripts/org-setup.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+const orgDisplayUserRes= JSON.parse(await $`sf org display user --target-org jsforce-test-org --json`)
+
+const userId= orgDisplayUserRes.result.id
+
+let csv ='OwnerId' + EOL
+
+for (let i = 0; i <= 5000; i++) {
+  csv += userId + EOL
+}
```
cristiand391, line 39: see: https://github.com/jsforce/jsforce/blob/2.0/test/bin/org-setup-sfdx.sh

scripts/org-setup.mjs
```diff
@@ -0,0 +1,84 @@
...
+await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`
+await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`
+
+const pushTopicValues = "Name='JSforceTestAccountUpdates' Query='SELECT Id, Name FROM Account' ApiVersion='54.0' NotifyForFields=Referenced NotifyForOperationCreate=true NotifyForOperationUpdate=true NotifyForOperationDelete=false NotifyForOperationUndelete=false"
+
+await $`sf data create record --sobject PushTopic --values ${pushTopicValues} --target-org jsforce-test-org`
+
+if (!process.env.CI) {
+  console.log(`Run tests using this scratch org by appending SF_LOGIN_URL=${orgDisplayUserRes.result.instanceUrl} SF_USERNAME=${orgDisplayUserRes.result.username} SF_PASSWORD=${orgDisplayUserRes.result.password}`)
+}
```
cristiand391, line 56: if running locally, print scratch org creds

scripts/org-setup.mjs
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Function examines the env var SF_OAUTH2_JWT_KEY to determine if it needs to be
+ * reformatted so when saved to a file the RSA key file contents are formatted
+ * properly.
+ *
+ * Throws an error if function is called and the env var is undefined
+ *
+ * returns a string that complies with RSA private key file format
+ */
+function formatJwtKey() {
```
cristiand391, line 67: copied from testkit to support setting the jwtKey as an env var:
https://github.com/salesforcecli/cli-plugins-testkit/blob/88dae2ce6afbe1b29ba3ff4a47fcd03ed399d3a0/src/hubAuth.ts#L28

test/connection-session.test.ts
```diff
@@ -5,6 +5,10 @@ import authorize from './helper/webauth';
 import config from './config';
 import { isNodeJS } from './helper/env';
 
+if (typeof jest !== 'undefined') {
+  jest.retryTimes(2);
+}
+
```
cristiand391, line 11: these auth tests launching a chromium instance sometimes fail (can't find the right HTML elements in the login page), rerunning it a few times help.

the `typeof` check is needed because this same file is loaded in a browser when running browser tests.

test/package/JSforceTestSuite/package.xml
```diff
@@ -61,5 +61,13 @@
...
\ No newline at end of file
+    <types>
+        <members>Admin</members>
+        <name>Profile</name>
+    </types>
+    <types>
+        <members>OauthOidc</members>
+        <name>Settings</name>
+    </types>
+    <version>58.0</version>
```
cristiand391, line 72: include profile and Oauth settings to enable oauth username-password flow.
This is needed because:
1) new orgs after Summer 23 release have this oauth flow disabled by default
2) test connection uses username-password flow:
https://github.com/jsforce/jsforce/blob/796d828fb4f6905f332bf0752fd66143f220e570/test/helper/connection-manager.ts#L39
**NOTE**: we talked about switching this to use the jwt flow but these tests files are also loaded in browser and `jsonwebtoken` is a node-only pkg.
3) need to set IP range 0.0.0.0 to 255.255.255.255 to allow using a only the password of the scratch org without the security token.

test/package/JSforceTestSuite/settings/OauthOidc.settings
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<OauthOidcSettings xmlns="http://soap.sforce.com/2006/04/metadata">
+    <blockOAuthUnPwFlow>false</blockOAuthUnPwFlow>
+    <blockOAuthUsrAgtFlow>true</blockOAuthUsrAgtFlow>
+    <oAuthCdCrdtFlowEnable>false</oAuthCdCrdtFlowEnable>
```
cristiand391, line 5: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_oauthoidcsettings.htm

scripts/org-setup.mjs
mshanemc: FYI, this code was because circle didn't handle the multiline string very well.

did you have the same problem with envs in GHA?

.github/workflows/tests.yml
mshanemc: I saw this issue about that https://github.com/browser-actions/setup-chrome/issues/216

mshanemc: Maybe this is an unusual time, but if not, would it make sense to put those urls/commit refs as GHA vars so it's easy to change them later?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,155 @@
...
+  #
+  #         echo "chromedriver_version=$CHROMEDRIVER_VERSION" >> "$GITHUB_ENV"
+  #     - uses: nanasess/setup-chromedriver@v2
```
mshanemc, line 25: pin 3rdParty actions to a SHA in case someone evil gets control of the repo

mshanemc (commented):
am fine taking this as is, as long as you've got WI in the epic for getting windows and browser tests working.

scripts/org-setup.mjs
cristiand391: > did you have the same problem with envs in GHA?

Ah, didn't know that. I haven't tried multiline.

.github/workflows/tests.yml
cristiand391: done

test/helper/webauth/index.ts
```diff
@@ -66,7 +66,7 @@ export default async function authorize(
         request.continue();
       }
     });
-    await page.goto(url, { waitUntil: 'networkidle2' });
+    await page.goto(url);
```
cristiand391, line 69: ran oauth tests locally with `headless: false` set here to launch chromium:
https://github.com/jsforce/jsforce/blob/796d828fb4f6905f332bf0752fd66143f220e570/test/helper/webauth/index.ts#L50

with `waitUntil: 'networkidle2'` it wasn't waiting properly for the login page to load. The default value (`load`) works fine.
https://pptr.dev/api/puppeteer.page.goto#remarks

mshanemc (approved):

PR #1356, opened 2023-07-25
chore: update node-fetch to 3.3.2

lukecotter:
This is to reduce bundle size.
It no longer has a dependency on `whatwg-url`
so overall bundle size is reduced.

bundlephobia thinks the dependecies of `node-fetch` will go from 283kB minified to 65.3 kB minified.
This is dependencies only and therefore excludes the size of node-fetch only, excluding deps.

I have also opened a PR for inquirer #SBoudrias/Inquirer.js#1271 which will also reduce the lodash dependency size (when bundled). So we can get that updated once merged too.
 
So once these are merged and the deps updated the overall size of jsforce can be reduced quite a bit.
#SBoudrias/Inquirer.js#1271
#auth0/node-jsonwebtoken#916

**[@salesforce/core](https://github.com/forcedotcom/sfdx-core)**
There is another large dep on lodash in `jsonwebtoken` but it looks like there is a PR already to address that too #auth0/node-jsonwebtoken#916. This is used by #forcedotcom/sfdx-core but it seems they have since bumped jsonwebtoken to 9.0.3 https://github.com/forcedotcom/sfdx-core/commit/87428c20d27a90f888d3a2f0c0095c24f394afe1


cristiand391:
Closing this since we decided to stay on node-fetch v2 for a bit longer until node's fetch is stable, see:
https://github.com/jsforce/jsforce/issues/1399#issuecomment-1967422881

PR #1368, opened 2023-10-03
fix: bump inquirer

iowillhoit:
Bumps the `inquirer` package to include a new version of `rxjs` (v7).

- `rxjs` v7 is 50% smaller
  - The `jsforce` bundle will decrease by 5Mbs
- This will also allow the `@salesforce/cli` to dedup this dependency 
  - This will decrease the `sf` cli bundle by 17Mbs

From what I can tell, `inquirer` is only used in the jsforce cli. I was able to enter the repl and answer prompts using this new version of `inquirer`

[@W-13735995@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-13735995)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "form-data": "^4.0.0",
         "fs-extra": "^8.1.0",
         "https-proxy-agent": "^5.0.0",
-        "inquirer": "^7.0.0",
+        "inquirer": "^8.2.6",
         "multistream": "^3.1.0",
         "node-fetch": "^2.6.1",
         "open": "^7.0.0",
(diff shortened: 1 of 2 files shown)


cristiand391 (approved):

cristiand391 (changes requested):
I've QA'd and works good but I the only breaking change from v7 -> v8 is dropping support for node 10:
https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%408.0.0

unfortunately jsforce v2 still supports node >= 8 so it would be a breaking change.

We can make this change once the v3 branch is up but for the purposes of the WI (reducing CLI bundle size) it will also require moving everything to v3.

cristiand391:
merged into https://github.com/jsforce/jsforce/pull/1369

PR #1376, opened 2023-11-09
fix: handle html responses better

mshanemc:
sometimes when an org is gone (expired, deleted, etc) there's an html error page for all urls ex:
https://computing-connect-6970-dev-ed.scratch.my.salesforce.com/services/data/v50.0

jsforce would set the error.message as **the entire contents of the html** which obviously isn't ideal for displaying to users.

So this creates a more generic message and moves the actual html to the `content` property of the error, in case anyone really wanted that html.

[@W-14454889@](https://gus.lightning.force.com/a07EE00001e6yycYAA)


src/http-api.ts
```diff
@@ -274,7 +274,14 @@ export class HttpApi<S extends Schema> extends EventEmitter {
             errorCode: `ERROR_HTTP_${response.statusCode}`,
             message: response.body,
           };
-    return new HttpApiError(error.message, error.errorCode);
+
+    return error.message.includes('<html ')
+      ? new HttpApiError(
```
cristiand391, line 279: AFAIK there's no endpoint that's expected to return HTML but since we have access to the res headers here, can you make it check that `content-type === text/html` is true instead if searching for html tags?

Screenshot 2023-11-13 at 18 52 54

Also, do you think it's worth keeping the HTML body in the error? we could just remove it and `logger.debug` it here, if someone is ever interested they can just set `JSFORCE_LOG_LEVEL=debug`

mshanemc: In the case of the 420 org error page, it's a lot of salesforce-branded SVG plotting and js scripts.

I was thinking there might be some simple error pages that people were parsing today, such that logger wouldn't be enough.  Logging might be a good idea in addition to keeping the error?

I don't know.  Hard call.

mshanemc: pushing up the header-based check with logger

cristiand391 (approved):

cristiand391:
**QA notes**:

trying to `org open` a deleted scratch org:

✅ catches html responses and throw friendly error:
**BEFORE**
```
➜  plugin-org git:(main) ./bin/dev.js org open -o test-hyuz7sb9tmg3@example.com
Error (1): <!DOCTYPE HTML>
<html lang=en-US>
<head>
<meta charset=UTF-8>
<title>Error Page</title>
    <style>
        /*! normalize.css v3.0.2 | MIT License | git.io/normalize */

        html {
            font-family: sans-serif;
            -ms-text-size-adjust: 100%;
            -webkit-text-size-adjust: 100%
...
```

**AFTER**
```
➜  plugin-org git:(main) ./bin/dev.js org open -o test-hyuz7sb9tmg3@example.com
Error (1): HTTP response contains html content.  See error.content for the full html
response.
```

✅ HTML gets logged:
```
JSFORCE_LOG_LEVEL=DEBUG ./bin/dev.js org open -o test-hyuz7sb9tmg3@example.com 
...
DEBUG   [http-api]  <request> method=GET, url=https://fun-speed-8102-dev-ed.scratch.my.salesforce.com/services/data
DEBUG   [http-api]  elapsed time: 268 msec
DEBUG   [http-api]  <response> status=420, url=https://fun-speed-8102-dev-ed.scratch.my.salesforce.com/services/data
DEBUG   [http-api]  html response.body: <!DOCTYPE HTML>
<html lang=en-US>
<head>
<meta charset=UTF-8>
<title>Error Page</title>
    <style>
        /*! normalize.css v3.0.2 | MIT License | git.io/normalize */

        html {
            font-family: sans-serif;
            -ms-text-size-adjust: 100%;
            -webkit-text-size-adjust: 100%
        }
```

PR #1381, opened 2023-11-24
feat: set content-length header

cristiand391:
feat: automatically set `content-length` header on REST/SOAP requests.

We've been getting a few reports that some internal envs started to enforce this header on requests, adding support in jsforce as a safety measure.

## Tests
while working on this I found the HTTP module doesn't any UT, there's some coverage for session refresh in different e2e tests but we today we don't have an easy way to test this behavior, I'll add some tests for this when working on
[W-14580746: [jsforce]: increase HTTP module test coverage]
(https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001fXgwbYAC/view)

## QA

not really an easy way to QA this without getting one of those internal envs, but you can try running tests with this env var:
`JSFORCE_LOG_LEVEL=DEBUG` to get the body size added in each request (I tried adding to mock the logger to grab the value and compare it in e2e tests but couldn't get it to work on browser/tests (you can't use jest specific stuff).

@W-14177773@

### DIFF
--- a/package.json
+++ b/package.json
@@ -209,7 +209,7 @@
   "dependencies": {
     "@babel/runtime": "^7.23.1",
     "@babel/runtime-corejs3": "^7.23.1",
-    "@sindresorhus/is": "^6.1.0",
+    "@sindresorhus/is": "^4",
     "@types/node": "^18.15.3",
     "abort-controller": "^3.0.0",
     "base64url": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,9 @@
     },
     "test:node": {
       "command": "jest --forceExit",
-      "dependencies": ["build:node:cjs"],
+      "dependencies": [
+        "build:node:cjs"
+      ],
       "files": [
         "test/**/*.ts",
         "jest.setup.js",
@@ -207,6 +209,7 @@
   "dependencies": {
     "@babel/runtime": "^7.23.1",
     "@babel/runtime-corejs3": "^7.23.1",
+    "@sindresorhus/is": "^6.1.0",
     "@types/node": "^18.15.3",
     "abort-controller": "^3.0.0",
     "base64url": "^3.0.1",
--- a/src/http-api.ts
+++ b/src/http-api.ts
@@ -185,8 +185,10 @@ export class HttpApi<S extends Schema> extends EventEmitter {
 
     const bodySize = getBodySize(request.body, headers);
 
+    const cannotHaveBody = ['GET', 'HEAD', 'OPTIONS'].includes(request.method);
+
     if (
-      request.method === 'POST' &&
+      !cannotHaveBody &&
       !!request.body &&
       !headers['transfer-encoding'] &&
       !headers['content-length'] &&
--- a/src/http-api.ts
+++ b/src/http-api.ts
@@ -127,6 +127,25 @@ export class HttpApi<S extends Schema> extends EventEmitter {
         // when session refresh delegate is available
         if (this.isSessionExpired(response) && refreshDelegate) {
           await refreshDelegate.refresh(requestTime);
+          // remove the `content-length` header after token refresh
+          //
+          // SOAP requests include the access token their the body,
+          // if the first req had an invalid token and jsforce successfully
+          // refreshed it we need to remove the `content-length` header
+          // so that it get's re-calculated again with the new body.
+          //
+          // REST request aren't affected by this because the access token
+          // is sent via HTTP headers
+          //
+          //
+          // `_message` is only present in SOAP requests
+          if (
+            Object.hasOwn(request, '_message') &&
+            request.headers &&
+            request.headers['content-length']
+          ) {
+            delete request.headers['content-length'];
+          }
           return this.request(request);
         }
         if (this.isErrorResponse(response)) {
--- a/src/http-api.ts
+++ b/src/http-api.ts
@@ -194,6 +194,9 @@ export class HttpApi<S extends Schema> extends EventEmitter {
       !headers['content-length'] &&
       !!bodySize
     ) {
+      this._logger.debug(
+        `missing 'content-length' header, setting it to: ${bodySize}`,
+      );
       headers['content-length'] = String(bodySize);
     }
     request.headers = headers;
--- a/src/http-api.ts
+++ b/src/http-api.ts
@@ -127,22 +127,22 @@ export class HttpApi<S extends Schema> extends EventEmitter {
         // when session refresh delegate is available
         if (this.isSessionExpired(response) && refreshDelegate) {
           await refreshDelegate.refresh(requestTime);
-          // remove the `content-length` header after token refresh
-          //
-          // SOAP requests include the access token their the body,
-          // if the first req had an invalid token and jsforce successfully
-          // refreshed it we need to remove the `content-length` header
-          // so that it get's re-calculated again with the new body.
-          //
-          // REST request aren't affected by this because the access token
-          // is sent via HTTP headers
-          //
-          //
-          // `_message` is only present in SOAP requests
+          /* remove the `content-length` header after token refresh
+           *
+           * SOAP requests include the access token their the body,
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -207,6 +209,7 @@
   "dependencies": {
     "@babel/runtime": "^7.23.1",
     "@babel/runtime-corejs3": "^7.23.1",
+    "@sindresorhus/is": "^4",
```
cristiand391, line 212: there's already a v6:
https://github.com/sindresorhus/is/releases

but had to use v4 because it's the latest CJS release.

When using ESM modules via `await import` webpack creates a new chunk just for it instead of keeping it in the main bundle:
https://webpack.js.org/guides/code-splitting/#dynamic-imports

I spent a few hours trying to find some workarounds but couldn't get it to inline this into the main bundle so keeping it at v4 until I found a solution for this (will need to for node-fetch v3 (ESM): https://github.com/jsforce/jsforce/pull/1356)

src/http-api.ts
```diff
@@ -162,6 +182,23 @@ export class HttpApi<S extends Schema> extends EventEmitter {
...
+      !!request.body &&
+      !headers['transfer-encoding'] &&
+      !headers['content-length'] &&
+      !!bodySize
+    ) {
+      this._logger.debug(
+        `missing 'content-length' header, setting it to: ${bodySize}`,
+      );
+      headers['content-length'] = String(bodySize);
+    }
```
cristiand391, line 201: see: https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2:~:text=A%20user%20agent%20SHOULD,anticipate%20such%20a%0A%20%20%20body.

https://github.com/sindresorhus/got/blob/3822412385506a1efef6580d270eae14086b9b43/source/core/index.ts#L618-L629

src/soap.ts
```diff
@@ -233,6 +234,21 @@ export class SOAP<S extends Schema> extends HttpApi<S> {
...
+      request.headers &&
+      !request.headers['transfer-encoding'] &&
+      !request.headers['content-length'] &&
+      !!bodySize
+    ) {
+      this._logger.debug(
+        `missing 'content-length' header, setting it to: ${bodySize}`,
+      );
+      request.headers['content-length'] = String(bodySize);
+    }
```
cristiand391, line 251: same as this but for SOAP requests:
https://github.com/jsforce/jsforce/pull/1381/files#r1412264591

one difference here is that this only checks if method is `POST`
[while GET is possible](https://stackoverflow.com/a/26339467), I haven't seen any examples of using non-POST in API docs. We can revert it later if it's not true.

src/util/get-body-size.ts
```diff
@@ -0,0 +1,42 @@
...
+    return body.getLengthSync();
+  }
+
+  return undefined;
+}
+
+type FormData = {
+  getBoundary: () => string;
+  getLength: (callback: (error: Error | null, length: number) => void) => void;
+} & Readable;
```
cristiand391, line 42: https://github.com/sindresorhus/got/blob/main/source/core/utils/get-body-size.ts

src/util/get-body-size.ts
```diff
@@ -0,0 +1,42 @@
...
+import { HttpBody } from '../types';
+import is from '@sindresorhus/is';
+
+export function getBodySize(
+  body: HttpBody | undefined,
+  headers: { [name: string]: string } | undefined,
+): number | undefined {
+  function isFormData(body: unknown): body is FormData {
+    return is.nodeStream(body) && is.function_((body as FormData).getBoundary);
+  }
```
cristiand391, line 11: this looks specifically for [this `FormData`](https://github.com/form-data/form-data), not the [web native](https://developer.mozilla.org/en-US/docs/Web/API/FormData) or [node native](https://nodejs.org/dist/latest-v20.x/docs/api/globals.html#class-formdata) versions.

we use `FormData` in jsforce.

src/util/get-body-size.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  if (is.buffer(body)) {
+    return body.length;
+  }
+  if (isFormData(body)) {
+    return body.getLengthSync();
+  }
+
+  return undefined;
+}
```
cristiand391, line 37: the only case not handled here is readable streams:
https://github.com/jsforce/jsforce/blob/5369d1d3e24bba0e384027d309c441c9ac5fa956/src/types/common.ts#L17

got stopped handling this in v12 too, see:
https://github.com/sindresorhus/got/issues/1486
https://github.com/sindresorhus/got/pull/1510

src/http-api.ts
```diff
@@ -126,6 +127,25 @@ export class HttpApi<S extends Schema> extends EventEmitter {
...
+          // SOAP requests include the access token their the body,
+          // if the first req had an invalid token and jsforce successfully
+          // refreshed it we need to remove the `content-length` header
+          // so that it get's re-calculated again with the new body.
+          //
+          // REST request aren't affected by this because the access token
+          // is sent via HTTP headers
+          //
+          //
+          // `_message` is only present in SOAP requests
```
mshanemc, line 141: style nit: I prefer `/**/` style for more than like 2 lines of comments

src/http-api.ts
```diff
@@ -126,6 +127,25 @@ export class HttpApi<S extends Schema> extends EventEmitter {
...
+          if (
+            Object.hasOwn(request, '_message') &&
+            request.headers &&
+            request.headers['content-length']
```
mshanemc, line 145: '_message' in request && request.headers?.['content-length']

src/http-api.ts
```diff
@@ -162,6 +182,23 @@ export class HttpApi<S extends Schema> extends EventEmitter {
...
+
+    const bodySize = getBodySize(request.body, headers);
+
+    const cannotHaveBody = ['GET', 'HEAD', 'OPTIONS'].includes(request.method);
+
+    if (
+      !cannotHaveBody &&
+      !!request.body &&
+      !headers['transfer-encoding'] &&
+      !headers['content-length'] &&
```
mshanemc, line 194: if 'content-length' is set to 0, does this still do what you want?

src/soap.ts
```diff
@@ -233,6 +234,21 @@ export class SOAP<S extends Schema> extends HttpApi<S> {
...
   beforeSend(request: HttpRequest & { _message: object }) {
     request.body = this._createEnvelope(request._message);
+
+    const bodySize = getBodySize(request.body, request.headers);
+
+    if (
+      request.method === 'POST' &&
+      request.headers &&
+      !request.headers['transfer-encoding'] &&
+      !request.headers['content-length'] &&
```
mshanemc, line 244: same question here about 0

src/util/get-body-size.ts
```diff
@@ -0,0 +1,42 @@
+import type { Readable } from 'node:stream';
+import { HttpBody } from '../types';
+import is from '@sindresorhus/is';
+
+export function getBodySize(
```
mshanemc, line 5: +1 for using an exported function.  It should be simple to write some true UT, which jsforce lacks, for this.

src/soap.ts
```diff
@@ -233,6 +234,21 @@ export class SOAP<S extends Schema> extends HttpApi<S> {
   /** @override */
   beforeSend(request: HttpRequest & { _message: object }) {
```
mshanemc, line 235: I'm not a fan of this design (pass an object for mutation).  I know you didn't start it.

But if http-api isn't exported and is only used via the 2 classes that extend it (soap and bulk?) maybe it's possible to fix it?

mshanemc: Or maybe that's too big and it's easier to stay in line with what jsforce is already doing?

mshanemc (approved):

mshanemc:
QA: created a record via `sf data create record` (uses `POST`) and JSFORCE_LOG_LEVEL=DEBUG

```
missing 'content-length' header, setting it to: 18
``` 

Seems fine

Issue #1374, opened 2023-11-07
node-only release


cristiand391:
## Current state
`jsforce` is an isomorphic (works on browser and node) library, as such it needs to ship multiple builds to cover most scenarios:

1) CJS & ESM build for node
2) ESM browser builds 
3) browser bundles + minified builds
4) TS source code (`src` folder)

For browser usage this isn't much of an issue since it either loads jsforce from a CDN (can get the full minified build only) or bundle it, but in node projects you end up with around ~28mb of unused code (`dist`, `browser` and `src` folders) + additional deps from the jsforce CLI and browser builds (`commander`, `inquirer`,  `open`, `@babel/runtime`, `@babel/runtime-corejs3`, `core-js`, etc).

To help with this, jsforce v3 will start publishing a new node-only pkg for node projects using jsforce that want a smaller package, this should help reducing the final build size and also during security scans (secury reports about library never used in node build).

### How it will work:
Each `jsforce` release will be followed by a `@jsforce/jsforce-node` release, both having the same package version (e.g. `jsforce@3.5.1` -> `@jsforce/jsforce-node@3.5.1`).
This new pkg will not live in a new repo or branch, it will be generated by modifying the package.json before publish (or locally for testing), specifically:
1. Changing `name` field to `@jsforce/jsforce-node`
2. Removing browser/CLI-related folders from `files` field.
3. Removing `bin` field
4. Removing browser/CLI dependencies

### Will I still be able to use `jsforce` in node projects?
Yes! `jsforce` will continue to work on browsers and node.

### Will this new pkg include additional, node-only features?
No. `@jsforce/jsforce-node` will only be a lighter(size) version of `jsforce`.

ref:
https://github.com/jsforce/jsforce/issues/534
https://github.com/jsforce/jsforce/issues/1117

fedozh:
Hi @cristiand391 , when new versions will be available on npm?

jaagaard01:
@cristiand391 hey was curious if the node only version has been released? I upgraded to the beta and havent been able to get conn.authorize to work but was fine before upgrade. Hoping the node only version fixes this issue for me. Thank you!

cristiand391:
@jaagaard01 yep, we published the first version a few hours ago:
https://www.npmjs.com/package/@jsforce/jsforce-node?activeTab=versions

note that the `0.0.1` is just a dummy package we published to save the name on npm, `3.0.0-next.1` is the valid one:
[...]

PR #1383, opened 2023-12-06
chore: update inquirer to inquirer/prompts

lukecotter:
This is related to #1356 

Updating to #SBoudrias/inquirer.js prompts will reduce the dep size from ~242kb to ~3kb.
It has been [written for smaller size and better performance](https://github.com/SBoudrias/Inquirer.js/discussions/1214)

An alternative to adopting inquirer prompts would be to update to inquirer v9 which removes a dependency on the whole of lodash #SBoudrias/Inquirer.js/pull/1271


lukecotter:
> fyi @lukecotter there's a `jsforce-node` package that don't include these CLI deps:
> 
> https://github.com/jsforce/jsforce/issues/1374#issuecomment-2229228522

Thanks for letting me know. I am using this via one of the sf cli libraries, the Salesforce-apex library. I see this now uses the jsforce-node version, I will update my deps for the smaller size, although last time I tried I think I had bundling issues.

If you don't mind I will point this at develop and sort any issues, this update will have benefits for the browser and other build types so I think it is still worth doing.

lukecotter:
Moved to #1671

PR #1391, opened 2024-01-23
test(streaming): fix flaky e2e test

cristiand391:
moar test fixes.

### DIFF
--- a/scripts/org-setup.mjs
+++ b/scripts/org-setup.mjs
@@ -75,11 +75,6 @@ await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --
 await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`;
 await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`;
 
-const pushTopicValues =
-  "Name='JSforceTestAccountUpdates' Query='SELECT Id, Name FROM Account' ApiVersion='54.0' NotifyForFields=Referenced NotifyForOperationCreate=true NotifyForOperationUpdate=true NotifyForOperationDelete=false NotifyForOperationUndelete=false";
-
-await $`sf data create record --sobject PushTopic --values ${pushTopicValues} --target-org jsforce-test-org`;
-
 if (!process.env.CI) {
   console.log(
     `Run tests using this scratch org by appending SF_LOGIN_URL=${orgDisplayUserRes.result.instanceUrl} SF_USERNAME=${orgDisplayUserRes.result.username} SF_PASSWORD=${orgDisplayUserRes.result.password}`,
--- a/test/bulk2.test.ts
+++ b/test/bulk2.test.ts
@@ -282,6 +282,8 @@ it('should call bulk api from invalid session conn with refresh fn, and return r
     },
   });
 
+  conn2.bulk2.pollTimeout = 90000;
+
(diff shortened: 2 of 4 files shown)


scripts/org-setup.mjs
```diff
@@ -75,11 +75,6 @@ await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --
 await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`;
 await $`sf force data bulk upsert --file ${bigTableCsv} --sobject BigTable__c --external-id Id --target-org jsforce-test-org`;
 
-const pushTopicValues =
-  "Name='JSforceTestAccountUpdates' Query='SELECT Id, Name FROM Account' ApiVersion='54.0' NotifyForFields=Referenced NotifyForOperationCreate=true NotifyForOperationUpdate=true NotifyForOperationDelete=false NotifyForOperationUndelete=false";
-
-await $`sf data create record --sobject PushTopic --values ${pushTopicValues} --target-org jsforce-test-org`;
-
```
cristiand391, line 82: This is used by the streaming e2e test, moved there so that the query looks for a specific account name.
This should help when tests are running on parallel, with the previous query it could get an `Account` record from other tests and now will just look for the account name that has an ID for the streaming test.

test/bulk2.test.ts
```diff
@@ -282,6 +282,8 @@ it('should call bulk api from invalid session conn with refresh fn, and return r
     },
   });
 
+  conn2.bulk2.pollTimeout = 90000;
+
```
cristiand391, line 286: when calling `loadAndWait` in L291 it was using the default 10s timeout for polling:
https://github.com/jsforce/jsforce/actions/runs/7628707431/job/20780447068#step:8:233

set to 90s like in other tests.

test/bulk2.test.ts
```diff
@@ -302,6 +304,9 @@ it('should call bulk api from invalid session conn without refresh fn, and raise
     logLevel: config.logLevel,
     proxyUrl: config.proxyUrl,
   });
+
+  conn3.bulk2.pollTimeout = 90000;
+
```
cristiand391, line 309: ditto

test/helper/webauth/index.ts
```diff
@@ -48,7 +48,7 @@ export default async function authorize(
   password: string,
 ) {
   const browser = await puppeteer.launch({
-    headless: true,
+    headless: 'new',
```
cristiand391, line 51: use new chrome headless mode:
https://github.com/jsforce/jsforce/actions/runs/7628707431/job/20780447068#step:8:210

test/streaming.test.ts
```diff
@@ -25,14 +25,28 @@ if (isNodeJS()) {
    */
   it('should subscribe to topic, create account, and receive event of account has been created', async () => {
     type Account = { Id: string; Name: string };
+
+    const id = Date.now();
+
+    const accountName = `My New Account #${id}`;
+
+    const pushTopicName = `Topic-${id}`;
```
cristiand391, line 33: unique `PushTopic` name to avoid duplicate error on test retries (needed now that the push topic is created inside this test).

We try to delete it at the end of the test but it might not if there's a failure before (and we could use a cleanup test helper from jest but these test run on browser so we can't use any jest-specific feature.

scripts/org-setup.mjs
cristiand391: happened here:
https://github.com/jsforce/jsforce/actions/runs/7617138418/job/20745503850?pr=1390#step:8:665

PR #1395, opened 2024-02-06
Add option to prevent cacheing SFDC responses

rdietrick:
Adds a noCache option to the connection which prevents browsers from cacheing SFDC responses.

When `conn.noCache == true`, a query string parameter (timestamp) is added to SFDC request URIs to prevent browsers from caching the requests.  

I'm building a browser-based application that allows the creation and modification of custom objects.  Without this option, describe calls don't return new fields that have been added to a custom object.


cristiand391:
closing this b/c v1 is now deprecated

PR #1397, opened 2024-02-08
feat: bulk2 refactor

cristiand391:
bulk2 module refactor:

**BREAKING CHANGES**:

* mode bulk2 class/types to a new `bulk2.ts` file
* `bulk2.query` returns a record stream (duplex) instead of an array of records
* `poll` method on query/ingest jobs throw on `Failed` state job instead of emitting `error` event.
* removes `queryJob.getResults` method. Can use `queryJob.result` to get a record stream if you are manually handling creating job/polling.

**Others**:
* add logging
* can set `columnDelimiter` and `lineEnding` when creating a query job.

## Rollout plan

Once this is merged and published I'll open a PR to bump jsforce v2 to jsforce-node v3 in sfdx-core, then get a pre-release to test on plugins (most should be fine except plugin-data).

## TODO

* e2e tests to cover streaming API
* allow to pick branch for cli NUTs

@W-14996843@


src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+    | 'CARET'
+    | 'COMMA'
+    | 'PIPE'
+    | 'SEMICOLON'
+    | 'TAB';
+  numberRecordsProcessed: string;
+  retries: string;
+  totalProcessingTime: string;
+  isPkChunkingSupported?: boolean;
+};
```
cristiand391, line 55: this used to be one type for both query and ingest jobs, there's some overlap but for instance states are different and also have unique props.

query job info:
https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_get_one_job.htm

ingest job info:
https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/get_job_info.htm

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+      connection: this.connection,
+      jobInfo: options,
+      pollingOptions: this,
+    });
+  }
+
+  /**
+   * Create, upload, and start bulkload job
+   */
+  async loadAndWaitForResults(
```
cristiand391, line 254: this name was chosen over `load` here:
https://salesforce.quip.com/CcboAPmPbgYb#TZAADAy1B6x
https://github.com/jsforce/jsforce/pull/1251#pullrequestreview-1062415091

await, promise gets resolved:
Screenshot 2024-02-14 at 15 29 10

if you don't await you get the Batch (writable stream):
Screenshot 2024-02-14 at 15 29 52

which you can write csv into (record stream system, there's an internal csv converter):
Screenshot 2024-02-14 at 15 45 03

cristiand391: a few questions:

1) do we still want to call it `loadAndWaitForResults`? I would be fine calling it `load`, even if it doesn't offer the same API as `conn.bulk.load`.

2) should it support returning a writable stream like `conn.bulk` that you can pipe csv into?
`loadAndWaitForResults` already accepts a readable input.

I didn't implement it because bulk2 requires the job to be closed for the data to be processed, so it would require some event handling to mark the job as "upload complete" once the data is written.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+   * @returns {RecordStream} - Record stream, convertible to a CSV data stream
+   */
+  async query(
+    soql: string,
+    options?: Partial<BulkV2PollingOptions> & {
+      scanAll?: boolean;
+      columnDelimiter?: QueryJobInfoV2['columnDelimiter'];
+      lineEnding?: QueryJobInfoV2['lineEnding'];
+    },
+  ): Promise<Parsable<Record>> {
```
cristiand391, line 297: `bulk2.query` now returns a duplex, consumers should listen to the record event and handle records there:
https://jsforce.github.io/document/#bulk-query

Unlike `conn.bulk`, you need to `await conn.bulk2.query`.

`conn.bulk` has an async IIFE: https://github.com/jsforce/jsforce/blob/588c6cf986650fb68f9b41314d5c48a9a5c5d478/src/api/bulk.ts#L969, bulk2 doesn't.

Check the e2e test below to see how it collects all records from the stream.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+            throw new Error('Job has been aborted');
+          case 'UploadComplete':
+          case 'InProgress':
+            await delay(interval);
+            break;
+          case 'Failed':
+            // unlike ingest jobs, the API doesn't return an error msg:
+            // https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_get_one_job.htm
+            this.logger.debug(res);
+            throw new Error('Query job failed to complete');
```
cristiand391, line 444: I put the response in debug logs but users will need to check in their org if three's more info about a query job failre.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+        method: 'GET',
+        path: this.locator
+          ? `${resultsPath}?locator=${this.locator}`
+          : resultsPath,
+        headers: {
+          Accept: 'text/csv',
+        },
+      });
+
+      resPromise.stream().pipe(resultDataStream);
```
cristiand391, line 507: this will attach the writable data stream to the readable inside the HTTP API module:
https://github.com/jsforce/jsforce/blob/588c6cf986650fb68f9b41314d5c48a9a5c5d478/src/http-api.ts#L63-L66

cristiand391: how it works:
https://github.com/jsforce/jsforce/blob/2.0/src/util/promise.ts

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+          case 'Aborted':
+            throw new Error('Job has been aborted');
+          case 'UploadComplete':
+          case 'InProgress':
+            await delay(interval);
+            break;
+          case 'Failed':
+            this.logger.debug(res);
+            throw new Error(
+              `Ingest job failed to complete due to: ${res.errorMessage}`,
```
cristiand391, line 723: Example of an error message (this is from last year when we spent a few days with Pete trying to figure out a NUT failure on windows because we didn't listen to the err event. v3 will just throw)

Screenshot (17)

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+        input: Record[] | Readable | string;
+      },
+  ): Promise<IngestJobV2Results<S>> {
+    if (!options.pollTimeout) options.pollTimeout = this.pollTimeout;
+    if (!options.pollInterval) options.pollInterval = this.pollInterval;
+
+    const job = this.createJob(options);
+    try {
+      await job.open();
+      await job.uploadData(options.input);
```
mshanemc, line 266: these might be good places for loggers to write things

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+   * @param interval Polling interval in milliseconds
+   * @param timeout Polling timeout in milliseconds
+   * @returns {Promise<void>} A promise that resolves when the job finishes successfully
+   */
+  async poll(
+    interval: number = this.pollingOptions.pollInterval,
+    timeout: number = this.pollingOptions.pollTimeout,
+  ): Promise<void> {
+    const jobId = getJobId(this.jobInfo);
+    const startTime = Date.now();
```
mshanemc, line 701: nit: compute the end time once, instead of computing it on every iteration

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+        const res = await this.check();
+        switch (res.state) {
+          case 'Open':
+            throw new Error('Job has not been started');
+          case 'Aborted':
+            throw new Error('Job has been aborted');
+          case 'UploadComplete':
+          case 'InProgress':
+            await delay(interval);
+            break;
```
mshanemc, line 719: this is a good place for, at a minimum, a logger statement about the progress.

I think it'd be cooler to `emit` some event, like 

```ts
numberRecordsProcessed?: number;
numberRecordsFailed?: number;
```

Imagine wanting to build a progress bar for the upload?

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+      this.once('response', () => resolve());
+      this.once('error', reject);
+    });
+
+    if (is.nodeStream(input)) {
+      // if input has stream.Readable interface
+      input.pipe(this.dataStream);
+    } else {
+      if (Array.isArray(input)) {
+        for (const record of input) {
```
mshanemc, line 948: I don't like that this mutates the input.  Imagine being a jsforce consumer, and you passed in your options, passed down through a bunch of stuff, and we changed all the booleans to string on your data.

mshanemc: maybe construct a new object that gets written from `Object.entries(Object.fromEntries(record).map("<boolean=>string stuff>"))

src/api/bulk2.ts
cristiand391: > Does Readable the readable input allow the readable side of a duplex/pipe? Like, obviously fs.createReadStream could be input here from a local file. What about a transform stream or pipe where someone wants to

yes, you could use a transform stream as long as you transform it back to plain csv.
https://github.com/jsforce/jsforce/blob/3.0/src/record-stream.ts#L235

> What's the use case for writeable?

consumers can start writing data early instead of having `loadAndWaitForResults` iterate over the array and write them individually.

here's an example (you can choose which Bulk API to use in some CRUD methods, this is how data is handled internally): https://github.com/jsforce/jsforce/blob/e760c2843d8d23fd57331661699dc538784f7a1c/src/query.ts#L950-L953

src/api/bulk2.ts
cristiand391: sounds good. There are a few async methods there, so they would match bulk2 👍🏼

src/api/bulk.ts
```diff
@@ -18,9 +17,8 @@ import {
   HttpResponse,
   Record,
   Schema,
-  Optional,
 } from '../types';
-import { isFunction, isObject } from '../util/function';
+import is from '@sindresorhus/is';
```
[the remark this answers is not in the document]
cristiand391: created [W-15061342](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001kIXxzYAG/view) to cover this.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+   * Deletes a query job.
+   */
+  async delete(): Promise<void> {
+    return this.createQueryRequest<void>({
+      method: 'DELETE',
+      path: `/${getJobId(this.jobInfo)}`,
+    });
+  }
+
+  private createQueryRequest<T>(request: BulkRequest) {
```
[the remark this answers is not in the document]
cristiand391: no. The baseUrl is set to `/services/data/vXX.X/jobs/query` but `request` can have a `path` that's appended to the baseUrl.

It's used in `queryJob.result` to paginate get query job results.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+> extends EventEmitter {
+  readonly connection: Connection<S>;
+  readonly logger: Logger;
+  readonly pollingOptions: BulkV2PollingOptions;
+  readonly jobData: JobDataV2<S, Opr>;
+  bulkJobSuccessfulResults: IngestJobV2SuccessfulResults<S> | undefined;
+  bulkJobFailedResults: IngestJobV2FailedResults<S> | undefined;
+  bulkJobUnprocessedRecords: IngestJobV2UnprocessedRecords<S> | undefined;
+  error: Error | undefined;
+  jobInfo: Partial<JobInfoV2>;
```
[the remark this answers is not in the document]
cristiand391: oh, missed these, I made these private in queryjob.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+   * Create a job representing a bulk operation in the org
+   *
+   * @returns {Promise<QueryJobInfoV2>} job information.
+   */
+  async open(): Promise<Partial<JobInfoV2>> {
+    try {
+      this.jobInfo = await this.createIngestRequest<JobInfoV2>({
+        method: 'POST',
+        body: JSON.stringify({
+          assignmentRuleId: this.jobInfo?.assignmentRuleId,
```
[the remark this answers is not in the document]
cristiand391: right, I think it was added because of type for `options.jobInfo` in the constructor, I'll update it.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+    this.jobInfo = options.jobInfo;
+    this.jobData = new JobDataV2<S, Opr>({
+      createRequest: (request) => this.createIngestRequest(request),
+      job: this,
+    });
+    // default error handler to keep the latest error
+    this.on('error', (error) => (this.error = error));
+  }
+
+  get id() {
```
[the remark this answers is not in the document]
cristiand391: It's a leftover from the previous implementation but now that `jobInfo` will be private we might need some getters/methods to expose these.

test/bulk2.test.ts
```diff
@@ -6,7 +6,8 @@ import ConnectionManager from './helper/connection-manager';
 import config from './config';
 import { isObject, isString } from './util';
 import { isNodeJS } from './helper/env';
-import { BulkOperation, IngestJobV2Results } from 'jsforce/lib/api/bulk';
+import { BulkOperation } from 'jsforce/lib/api/bulk';
```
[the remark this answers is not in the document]
cristiand391: I can't remember exactly if it's needed (don't think browser, jest gave me some problems a while ago and there's a module mapper so that `jsforce/` works).

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+  columnDelimiter:
+    | 'BACKQUOTE'
+    | 'CARET'
+    | 'COMMA'
+    | 'PIPE'
+    | 'SEMICOLON'
+    | 'TAB';
+  numberRecordsProcessed: string;
+  retries: string;
+  totalProcessingTime: string;
```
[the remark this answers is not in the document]
cristiand391: corrected, here's an example from an insert job:
```
DEBUG   [bulk2:IngestJobV2]  {
  id: '7508N0000020D8WQAU',
  operation: 'insert',
  object: 'Account',
  createdById: '0058N000005rF2eQAE',
  createdDate: '2024-02-16T21:55:12.000+0000',
  systemModstamp: '2024-02-16T21:55:14.000+0000',
  state: 'JobComplete',
  concurrencyMode: 'Parallel',
  contentType: 'CSV',
  apiVersion: 60,
  jobType: 'V2Ingest',
  lineEnding: 'LF',
  columnDelimiter: 'COMMA',
  numberRecordsProcessed: 3,
  numberRecordsFailed: 0,
  retries: 0,
  totalProcessingTime: 184,
  apiActiveProcessingTime: 113,
  apexProcessingTime: 53,
  isPkChunkingSupported: false
}
```

src/api/bulk2.ts
cristiand391: added logs to each job method instead of here. Couldn't think of any specifics to put here that wouldn't apply to used methods.

src/api/bulk2.ts
cristiand391: added `inProgress` event that include the full response.

bulk v1 was doing the same but emitting `progress`, now both to `inProgress`.

src/api/bulk2.ts
cristiand391: fixed by using `structuredClone` in both bulk modules.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1082 @@
...
+    >
+  >;
+
+type NewQueryJobOptions = {
+  query: string;
+  operation: QueryJobInfoV2['operation'];
+} & Partial<Pick<QueryJobInfoV2, 'columnDelimiter' | 'lineEnding'>>;
+
+type CreateIngestJobV2Options = {
+  bodyParams: NewIngestJobOptions;
```
cristiand391, line 109: I would have preferred `options` instead of `bodyParams` but this is already inside an `options` object that constructors take.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1082 @@
...
+   * Default: 1000 (1 second)
+   */
+  public pollInterval = 1000;
+
+  /**
+   * Polling timeout in milliseconds
+   *
+   * Default: 30000 (30 seconds)
+   */
+  public pollTimeout = 30000;
```
cristiand391, line 192: increased the default timeout to 30s (was 10s before) in both bulk modules.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1082 @@
...
+   */
+  public job(
+    type: 'query',
+    options: Pick<ExistingQueryJobV2Options, 'id'>,
+  ): QueryJobV2<S>;
+  public job(
+    type: 'ingest',
+    options: Pick<ExistingIngestJobOptions, 'id'>,
+  ): IngestJobV2<S>;
+  public job(
```
cristiand391, line 244: get an existing job by providing an id.

by default will get an ingest job (this matches bulkv1.job), can get a query job too.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+  lineEnding?: QueryJobInfoV2['lineEnding'];
+};
+export class QueryJobV2<S extends Schema> extends EventEmitter {
+  private readonly connection: Connection<S>;
+  private readonly logger: Logger;
+  private readonly body: QueryJobBody;
+  private readonly pollingOptions: BulkV2PollingOptions;
+  private error: Error | undefined;
+  private jobInfo: QueryJobInfoV2 | undefined;
+  private locator: Optional<string>;
```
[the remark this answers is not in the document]
cristiand391: doh, missed this. will push once tests pass

src/api/bulk2.ts
```diff
@@ -0,0 +1,1082 @@
...
+  private readonly logger: Logger;
+  private readonly _id?: string;
+  private readonly bodyParams?: NewQueryJobOptions;
+  private readonly pollingOptions: BulkV2PollingOptions;
+  private error: Error | undefined;
+  private jobInfo?: QueryJobInfoV2;
+  private locator: Optional<string>;
+
+  constructor(conn: Connection<S>, options: ExistingQueryJobV2Options);
+  constructor(conn: Connection<S>, options: CreateQueryJobV2Options);
```
cristiand391, line 361: overload constructor so that you provide either an `id` or `bodyParams` in options when creating a job instance.

src/api/bulk2.ts
```diff
@@ -0,0 +1,1000 @@
...
+          assignmentRuleId: this.jobInfo?.assignmentRuleId,
+          externalIdFieldName: this.jobInfo?.externalIdFieldName,
+          object: this.jobInfo?.object,
+          operation: this.jobInfo?.operation,
+          lineEnding: this.jobInfo?.lineEnding,
+        }),
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8',
+        },
+        responseType: 'application/json',
```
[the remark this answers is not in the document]
cristiand391: `responseType` is part of the `HttpApi` class, it's stored as a class prop to check the body later.

mshanemc (approved):

mshanemc:
QA continued

I found one more tiny edge case.  I can call job.getAllResults on an ingest job and get back this (note the unprocessedRecords is a string and not a [] .
```ts
{
  successfulResults: [],
  failedResults: [],
  unprocessedRecords: 'Id,\n"001O8000008hZzLIAU",\n"001O8000008hZzPIAU"'
} 
```
you can replicate this by sending it an invalid csv like (the commas on the EOL are bad)
```csv
Id,
"001O8000008hZzLIAU",
"001O8000008hZzPIAU"
```

so that we end up with a code-10 TypeError `results.unprocessedRecords.map is not a function`

PR #1403, opened 2024-03-15
feat: retry on network errors

cristiand391:
This PR adds a simple retry mechanism for the node-fetch wrapper.

Implements some options of undici's (node's fetch) retry handler:
https://undici.nodejs.org/#/docs/api/RetryAgent

jsforce v4 will target native fetch on node/browser so the retry API will stay the same for node consumers.

### Caveats:
1. this code will very likely get removed in jsforce v4 (node's fetch handles retries)
2. the retry mechanism is at the low-level request helper and relies on node-fetch actually throwing, 

### How is the node-fetch wrapper used?

Connection class is instantiated, it sets the HTTP transport:
https://github.com/jsforce/jsforce/blob/3a88038617494cd2c877f152363218f17939c400/src/connection.ts#L346-L350

then `HttpApi.httpRequest` gets the `request` wrapper here:
https://github.com/jsforce/jsforce/blob/3a88038617494cd2c877f152363218f17939c400/src/transport.ts#L50

the `request` func is part of the public API, will be doc'd here:
https://github.com/jsforce/jsforce-website/pull/10 (needs updates) 

@W-14990080@

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -11,6 +11,7 @@ import {
 } from './request-helper';
 import { HttpRequest, HttpRequestOptions } from './types';
 import { getLogger } from './util/logger';
+import is from '@sindresorhus/is';
 
 /**
  *
@@ -95,6 +96,13 @@ async function startFetchRequest(
         if (retryCount === maxRetry) return false;
 
         if (!retryOpts?.methods?.includes(request.method)) return false;
+
+        if (is.nodeStream(body) && Readable.isDisturbed(body)) {
+          logger.debug(
+            'Body of type stream was read, unable to retry request.',
+          );
+          return false;
+        }
 
         if (
           'code' in error &&
--- a/src/request.ts
+++ b/src/request.ts
@@ -68,7 +68,7 @@ async function startFetchRequest(
   ): Promise<Response> => {
     const fetchOpts: RequestInit = {
       ...rrequest,
-      ...(input && /^(post|put|patch)$/i.test(request.method)
+      ...(request.body && input && /^(post|put|patch)$/i.test(request.method)
         ? { body: input }
         : {}),
       redirect: 'manual',
--- a/src/request.ts
+++ b/src/request.ts
@@ -68,7 +68,7 @@ async function startFetchRequest(
   ): Promise<Response> => {
     const fetchOpts: RequestInit = {
       ...rrequest,
-      ...(request.body && input && /^(post|put|patch)$/i.test(request.method)
+      ...(input && /^(post|put|patch)$/i.test(request.method)
         ? { body: input }
         : {}),
       redirect: 'manual',
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events';
 import { Duplex, Readable, Writable } from 'stream';
-import fetch from 'node-fetch';
+import fetch, { Response, RequestInit, FetchError } from 'node-fetch';
 import AbortController from 'abort-controller';
 import createHttpsProxyAgent from 'https-proxy-agent';
 import {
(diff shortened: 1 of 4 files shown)


src/request.ts
```diff
@@ -34,25 +36,107 @@ async function startFetchRequest(
...
+      'EPIPE',
+    ],
+    methods: options.retry?.methods ?? [
+      'GET',
+      'PUT',
+      'HEAD',
+      'OPTIONS',
+      'DELETE',
+    ],
+  };
```
cristiand391, line 67: these are the default errorCodes and methods in node's fetch retry handler:
https://undici.nodejs.org/#/docs/api/RetryAgent?id=parameter-retryhandleroptions

plus these 2: `ETIMEDOUT` and `EPIPE`

src/request.ts
```diff
@@ -34,25 +36,107 @@ async function startFetchRequest(
...
+      const error = err as Error | FetchError;
+
+      // request was canceled by consumer (AbortController), skip retry and rethrow.
+      if (error.name === 'AbortError') {
+        throw error;
+      }
+
+      const shouldRetry = (): boolean => {
+        // only retry on operational errors
+        if (error.name != 'FetchError') return false;
```
cristiand391, line 95: see: https://github.com/node-fetch/node-fetch/blob/2.x/ERROR-HANDLING.md#error-handling-with-node-fetch

src/request.ts
```diff
@@ -34,25 +36,107 @@ async function startFetchRequest(
...
+        if (retryCount === maxRetry) return false;
+
+        if (!retryOpts?.methods?.includes(request.method)) return false;
+
+        if (is.nodeStream(body) && Readable.isDisturbed(body)) {
+          logger.debug(
+            'Body of type stream was read, unable to retry request.',
+          );
```
cristiand391, line 103: ensure request with a stream body aren't retried if the stream was consumed.

src/types/common.ts
```diff
@@ -36,6 +37,11 @@ export type HttpRequest = {
 };
 
 export type HttpRequestOptions = {
+  retry?: {
+    maxRetries?: number;
+    errorCodes?: string[];
+    methods?: HttpMethods[];
+  };
```
cristiand391, line 44: tried to match undici API: https://undici.nodejs.org/#/docs/api/RetryAgent?id=parameter-retryhandleroptions

it doesn't expose `statusCodes`/`retryAfter` because the retry mechanism relies on node-fetch throwing where it can't access status code or headers.

test/http-api.test.ts
```diff
@@ -16,6 +18,95 @@ describe('HTTP API', () => {
     loginUrl,
   });
 
+  describe('network retry', () => {
```
mdonnalley, line 21: Might be useful to have a test that tests that the retries stop after the max is reached

mdonnalley (approved):

mdonnalley:
# QA


**setup**
- checkout branch, build
- `yarn link jsforce` in sfdx-core
- `yarn link @salesforce/core` in plugin-org
- `sf plugins link` --no-install in plugin-org

:green_circle: `sf org create scratch --edition developer --alias my-org`

The following I believe are sufficiently covered by the unit tests (once #1405 is merged). The alternative would be to write a hook for `sf` that uses `nock` to intercept http requests and fail them a couple of times before succeeding. While that would be a decent test, it's essentially the same thing as what the unit tests are doing.

:green_circle: retries on supported errors
:green_circle: no retries on unsupported errors
:green_circle: retries on specified errors
:green_circle: retries on supported methods
:green_circle: no retries on unsupported methods
:green_circle: retries on specified methods
:green_circle: no more retries after max is reached

PR #1412, opened 2024-04-24
test: run e2e tests on windows

cristiand391:
Enable e2e tests on Windows, updated the org-setup script to fix some path issues that where blocking this.

After getting these running I was getting this failure even after setting the correct lineEnding value for windows:
https://github.com/jsforce/jsforce/actions/runs/8818324210/job/24206906041#step:8:645

turns out `bulk2.loadAndWaitForResults` wasn't passing job bodyParams down when creating the job, fixed in this commit:
https://github.com/jsforce/jsforce/commit/8e63fd024c8afe6cf6c2f721072727b91c94aee9

@W-13879927@

### DIFF
--- a/src/api/bulk2.ts
+++ b/src/api/bulk2.ts
@@ -270,15 +270,14 @@ export class BulkV2<S extends Schema> {
     if (!options.pollTimeout) options.pollTimeout = this.pollTimeout;
     if (!options.pollInterval) options.pollInterval = this.pollInterval;
 
-    const job = this.createJob({
-      object: options.object,
-      operation: options.operation,
-    });
+    const { pollInterval, pollTimeout, input, ...createJobOpts } = options;
+
+    const job = this.createJob(createJobOpts);
     try {
       await job.open();
-      await job.uploadData(options.input);
+      await job.uploadData(input);
       await job.close();
-      await job.poll(options.pollInterval, options.pollTimeout);
+      await job.poll(pollInterval, pollTimeout);
       return await job.getAllResults();
     } catch (error) {
       const err = error as Error;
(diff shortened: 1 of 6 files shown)


test/bulk2.test.ts
```diff
@@ -197,10 +197,11 @@ if (isNodeJS()) {
   it('should bulk insert from file and return inserted results', async () => {
     // insert 100 account records from csv file
     const csvStream = fs.createReadStream(
-      path.join(__dirname, 'data/Account_bulk2_test.csv'),
+      path.join(__dirname, 'data', 'Account_bulk2_test.csv'),
     );
 
     const res = await conn.bulk2.loadAndWaitForResults({
+      lineEnding: require('os').platform() === 'win32' ? 'CRLF' : 'LF',
```
cristiand391, line 204: default value is `LF`, need to be `CRLF` when working on windows or you get a `ClientInputError`.

plugin-data does the same here:
https://github.com/salesforcecli/plugin-data/blob/85b73121d52fa02fb840b8bd2d6838bdf35ac632/src/bulkOperationCommand.ts#L112

src/api/bulk2.ts
```diff
@@ -270,15 +270,14 @@ export class BulkV2<S extends Schema> {
     if (!options.pollTimeout) options.pollTimeout = this.pollTimeout;
     if (!options.pollInterval) options.pollInterval = this.pollInterval;
 
-    const job = this.createJob({
-      object: options.object,
-      operation: options.operation,
-    });
+    const { pollInterval, pollTimeout, input, ...createJobOpts } = options;
+
+    const job = this.createJob(createJobOpts);
```
cristiand391, line 275: see the `NewIngestJobOptions` type above, it has some properties that can be set by the user but weren't passed down to `createJob` (`lineEnding` was the one I needed for the tests to pass, see my other comment: https://github.com/jsforce/jsforce/pull/1412/files#r1578340035).

iowillhoit (approved):

PR #1416, opened 2024-05-06
Wr/eslintII

WillieRuemmele:
adds more eslint rules, fixes where reasonable

@W-15061342@


src/api/bulk2.ts
```diff
@@ -281,7 +282,7 @@ export class BulkV2<S extends Schema> {
       return await job.getAllResults();
     } catch (error) {
       const err = error as Error;
-      this.logger.error(`bulk load failed due to: ${err}`);
+      this.logger.error(`bulk load failed due to: ${err.name}`);
```
cristiand391, line 285:
```suggestion
      this.logger.error(`bulk load failed due to: ${err.message}`);
```

src/api/bulk2.ts
```diff
@@ -332,7 +333,7 @@ export class BulkV2<S extends Schema> {
       queryRecordsStream.pipe(dataStream);
     } catch (error) {
       const err = error as Error;
-      this.logger.error(`bulk query failed due to: ${err}`);
+      this.logger.error(`bulk query failed due to: ${err.name}`);
```
cristiand391, line 336:
```suggestion
      this.logger.error(`bulk query failed due to: ${err.message}`);
```

src/api/bulk2.ts
```diff
@@ -355,8 +356,7 @@ export class QueryJobV2<S extends Schema> extends EventEmitter {
   private jobInfo?: QueryJobInfoV2;
   private locator?: string;
 
-  constructor(conn: Connection<S>, options: ExistingQueryJobV2Options);
-  constructor(conn: Connection<S>, options: CreateQueryJobV2Options);
+  constructor(conn: Connection<S>, options: ExistingQueryJobV2Options | CreateQueryJobV2Options);
```
cristiand391, line 359: nice catch, it seems even this line can be removed since the constructor below has the same signature.

src/api/bulk2.ts
```diff
@@ -617,8 +617,7 @@ export class IngestJobV2<S extends Schema> extends EventEmitter {
   /**
    *
    */
-  constructor(conn: Connection<S>, options: ExistingIngestJobOptions);
-  constructor(conn: Connection<S>, options: CreateIngestJobV2Options);
+  constructor(conn: Connection<S>, options: ExistingIngestJobOptions| CreateIngestJobV2Options);
```
cristiand391, line 620: ditto, same signature

src/browser/request.ts
```diff
@@ -106,7 +106,7 @@ async function startFetchRequest(
     () => controller?.abort(),
   );
   const headers: { [key: string]: any } = {};
-  // @ts-expect-error
+  // @ts-expect-error no .keys()?
   for (const headerName of res.headers.keys()) {
```
cristiand391, line 110: I added this `ts-expect-error` when updating TS to include DOM types.
It's probably broken (unless TS types are wrong) but I left it as it is because no other code users this (even in e2e).

src/query.ts
```diff
@@ -247,7 +247,7 @@ export class Query<
         ? this.urlToLocator(locator)
         : locator;
     } else {
-      this._logger.debug(`config is QueryConfig: ${config}`);
+      this._logger.debug(`config is QueryConfig: ${JSON.stringify(config)}`);
```
cristiand391, line 250: nice, current logs:
```
DEBUG   [http-api]  <response> status=200, url=https://na40-dev-hub.my.salesforce.com/services/data
DEBUG   [query]  config is QueryConfig: [object Object]
```

src/request.ts
```diff
@@ -118,6 +118,7 @@ async function startFetchRequest(
         ? { body: input }
         : {}),
       redirect: 'manual',
+      // @ts-expect-error - differing types of signal? this started abruptly
```
cristiand391, line 121: yep, we use an AbortController polyfill:
https://github.com/mysticatea/abort-controller/issues/36

might drop it in v4 (use browser/node native AC).

src/request.ts
```diff
@@ -154,7 +155,7 @@ async function startFetchRequest(
 
       if (shouldRetryRequest(retryOpts.maxRetries, error)) {
         logger.debug(`retrying for the ${retryCount + 1} time`);
-        logger.debug(`Error: ${error}`);
+        logger.debug(`Error: ${JSON.stringify(error)}`);
```
cristiand391, line 158:
```suggestion
        logger.debug(`Error: ${error.message}`);
```

test/soap.test.ts
```diff
@@ -88,9 +88,10 @@ describe('convert and merge', () => {
       conn370.metadata.read('CustomField', 'Lead.Status'),
     ]);
     leadIds = rets.map((r) => r.id!);
-    convertedStatus = (statusField as any).picklist.picklistValues
-      .filter((pv: any) => pv.converted === 'true')
-      .map((pv: any) => pv.fullName)[0];
+    // @ts-expect-error  @typescript-eslint/no-unnecessary-type-assertion weird CI only linting?
+    convertedStatus = statusField.picklist.picklistValues
```
cristiand391, line 92: statusField was being casted before:
`(statusField as any)`

probably the outdated MDAPI types.

cristiand391 (changes requested):
left some comments/suggestions.

Can you remove the `dist` changes? those should be pushed on release (which doesn't seem to be happen for v3, I'll create a WI for that).

src/browser/request.ts
WillieRuemmele: and I just did my part of adding why we need the `expect-error` 🤷‍♂️

test/soap.test.ts
WillieRuemmele: yeah - it was throwing an "unnecessary type assertion" error [before] (https://github.com/jsforce/jsforce/actions/runs/8976679543/job/24653986013) with `as any`, but then in the editor it was passing

cristiand391 (approved):

Issue #1352, opened 2023-06-27
Non-standard login URLs break authentication


bluelhf:
# Issue Description

When the Salesforce instance URL is a subdomain of `lightning.force.com` (a perfectly valid instance URL), HTTP authentication breaks.

## Cause

This issue is caused by JSForce's use of the [`request`](https://www.npmjs.com/package/request) NPM package as a HTTP client, which **has been deprecated for over three years.**

1. Salesforce responds to requests to `lightning.force.com` with a <kbd>302 Found</kbd> redirect.
2. The `request` HTTP client breaks HTTP specification and loses the request method and body after **all non-307 redirects** .
3. Since it's forgot about the request method, `request` makes a GET request to the SOAP login endpoint, to which Salesforce responds with an error.

## Steps to reproduce

1. Set up a Salesforce instance.
2. Attempt to log in to the instance via JSForce using the `lightning.force.com` URL and appropriate credentials
3. See the error response from JSForce:
```html
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 405 Only POST allowed</title>
</head>
<body><h2>HTTP ERROR 405</h2>
<p>Problem accessing /services/Soap/u/42.0. Reason:
<pre>    Only POST allowed</pre></p><hr /><br/>
<!----- 8< snip ! 8< ----->
```
5. Cry

## Proposed Solution

In `v2`, JSForce will use `node-fetch`. Yay!

For now, a quick hack would be to rewrite the login URL domain from `lightning.force.com` to `my.salesforce.com`.

paustint:
Yes, [enhanced domains](https://help.salesforce.com/s/articleView?id=sf.domain_name_url_format_changes_enable_enhanced.htm&type=5) changed the url structure for accessing salesforce.

gmillerd:
I have never used our domains and instead used just login.salesforce.com and test.salesforce.com and have the login username do the work. Is there a gap in doing this?

paustint:
@gmillerd - there is no gap, but many orgs disable login through these URLs and require going through the custom domain.

And IMO, this ticket is invalid, the login URL for a domain is not the `.lightning` subdomain, it is either `login.salesforce.com` or `test.salesforce.com` or the actual instance domain.

Salesforce calls out the login domain explicitly in the docs https://help.salesforce.com/s/articleView?id=sf.domain_name_url_format_changes_enable_enhanced.htm&type=5

[...]

PR #1418, opened 2024-05-13
fix: throw error if lightning url passed as instance url

WillieRuemmele:
@W-15748051@
https://github.com/jsforce/jsforce/issues/1352

validates instanceUrls as non lightning before creating connections with them

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -321,6 +321,11 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
     } = config;
     this.loginUrl = loginUrl || defaultConnectionConfig.loginUrl;
     this.instanceUrl = instanceUrl || defaultConnectionConfig.instanceUrl;
+
+    if (this.isLightningInstance()) {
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -1599,6 +1604,14 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
    * Module which manages process rules and approval processes
    */
   process = new Process(this);
+
+  public isLightningInstance(): boolean {
```
cristiand391, line 1608: let's make this private, it doesn't take URLs (in that case it would probably an util function) and after you create a connection it's guaranteed it doesn't use a lightning domain.

cristiand391 (changes requested):
looks good, just one request to make the new method private.

cristiand391 (approved):

cristiand391:
**QA notes**:

✅ setting `instanceUrl` to a lightning URL makes it throw
✅ only `instanceUrl` is validated, can still use `(login|test).salesforce.com` as `loginUrl` in connection.

Issue #1385, opened 2023-12-06
SOQL query using autofetch=true drops a batch of results set when using queryMore after the initial query


peternhale:
I ran into an issue while using query, with autoFetch true and then wanting to fetch more records using queryMore.

As I understand autoFetch, once the query call finishes the response will contain the properties done, totalSize, records[] and optionally nextRecordsUrl. The SOQL query I am using returns more records than maxFetch, with the properties:
* done = false
* totalSize = number of records in my query
* records[maxFetch is size]
* nextRecordsUrl= URL to the next batch of 2000 records.

At this moment I use queryMore function to start getting the remaining records. What I am finding is that with autoFetch = true, is one batch will be missing and I believe it is the batch that follows the number of records specified by maxFetch.

For example:
* SOQL query returns 20,000 records
* maxFetch set to 10,000 records
* after following all presented nextRecordsUrls the total number of records is 18,000

I am using the most recent version of v2@beta and node 18.18.*

I wrote a reproducer (see [gist](https://gist.github.com/peternhale/37b6681535888ade20673c55ddc94537)) that displays the key properties from the query. Both use maxFetch = 2,000. The first is with autoFetch false and as you can see all the queried rows are returned. Each query/queryMore result displays the number of the query/queryMore, batch size and `nextRecordsUrl`. Notice the first `nextRecordsUrl` ends with `-2000`.

```
Executing query: SELECT Id, Name FROM Account with autoFetch: false
Query count: 1 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-2000
Query count: 2 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-4000
Query count: 3 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-6000
Query count: 4 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-8000
Query count: 5 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-10000
Query count: 6 Records in this batch: 2000 nextRecordsUrl: https://<redacted>/services/data/v50.0/query/0r8xx2PivsRiHU6ACN-12000
[...]

PR #1424, opened 2024-05-17
Cd/fix query missing batch

cristiand391:
https://github.com/jsforce/jsforce/issues/1385

### DIFF
--- a/test/query.test.ts
+++ b/test/query.test.ts
@@ -75,6 +75,19 @@ describe('big tables and autoFetch', () => {
     assert.ok(records.length === totalRecordCount);
   });
 
+  it('should return the first batch with autoFetch and maxFetch set to 2000', async () => {
+    let requestQty = 0
+
+    const result = await conn.query(bigTableQuery, { autoFetch: true, maxFetch: 2000}).on('fetch', () => {
+      requestQty++
+    })
+
+    expect(result.records.length).toBe(2000);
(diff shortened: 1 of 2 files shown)


src/query.ts
```diff
@@ -674,6 +674,7 @@ export class Query<
       this._finished ||
       data.done ||
       !autoFetch ||
+      this.records.length === maxFetch ||
```
cristiand391, line 677: `maxFetch` is always set (if user doesn't set it, it defaults to `10_000`). 

set finished=true if we already got enough records in the first request.

test/query.test.ts
```diff
@@ -75,6 +75,19 @@ describe('big tables and autoFetch', () => {
...
+    let requestQty = 0
+
+    const result = await conn.query(bigTableQuery, { autoFetch: true, maxFetch: 2000}).on('fetch', () => {
+      requestQty++
+    })
+
+    expect(result.records.length).toBe(2000);
+    expect(result.nextRecordsUrl).toBeTruthy();
+    expect(result.done).toBe(false);
+    expect(requestQty).toBe(1);
```
cristiand391, line 88: this ensures only 1 request was done.

shetzel (approved):

shetzel:
I see the correct output now using Pete's script, and the incorrect output without your fix.

PR #1499, opened 2024-06-06
Fix: Version information is not properly returned on JSforce ver 3.0

### LINKED ISSUE #1498: Version information is not properly returned on JSforce ver 3.0
From CLI, it returns wrong version information:

```shell
$ npx jsforce@3.2.0 --version
2.0.0-beta.8
```

As the above command refers the content of `src/VERSION.ts`, it should be maintained according to the update of version info in package.json.

Note: Formerly (in 1.0 era) it has "prepublish" script to confirm it is synced, so it was possible to notice if there is a difference before publishing as a package.
https://github.com/jsforce/jsforce/blob/c04515846e91f84affa4eb87a7b2adb1f58bf04d/package.json#L35

stomita:
Fixes #1498

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,8 @@
     "clean": "rimraf --glob lib/* module/* browser/*",
     "clean-test-failures-file": "rimraf test/karma/karma-failed-tests.txt",
     "jsforce-node": "ts-node scripts/jsforce-node.ts",
-    "jsforce-node:dev": "ts-node scripts/jsforce-node.ts --dev"
+    "jsforce-node:dev": "ts-node scripts/jsforce-node.ts --dev",
+    "prepublish": "ts-node -e \"if(process.env.npm_package_version!==require('./src/VERSION').default){console.error('The pacakge.json version is not matching to ./src/VERSION.js');process.exit(1)}\""
   },
   "files": [
     "README.md",
--- a/src/VERSION.ts
+++ b/src/VERSION.ts
@@ -1,1 +1,1 @@
-export default '2.0.0-beta.8';
+export default '3.2.0';


cristiand391:
@stomita we added a publish workflow that does the bump in `package.json` here:

https://github.com/jsforce/jsforce/actions/workflows/manualRelease.yml
Screenshot 2024-06-06 at 11 08 57 AM

https://github.com/jsforce/jsforce/blob/b661afdd21729812974f1056e392f2f58fc21096/.github/workflows/manualRelease.yml#L27C1-L36C25

it creates a GH release with that version and push a commit with the bump:

https://github.com/jsforce/jsforce/commit/10c79c51e422b9085d4f92b52214010acdf9bdbc

then the GH release trigger the npm-publish workflow:
https://github.com/jsforce/jsforce/blob/3.0/.github/workflows/onRelease.yml


If we merge this change it'll cause publish failures if someone guessed the wrong version according to semver (or forgot to do it).
What do you think if we bump the `package.json` **and** `src/VERSION.ts` files at the same time? the action we use to do the semver bump supports pre-commit hooks (JS files), so *I think* it's doable.
https://github.com/TriPSs/conventional-changelog-action?tab=readme-ov-file#pre-commit-hook

stomita:
@cristiand391 it would be ok if you can change to bump the version to both files. thanks.

AllanOricil:
Why not make version be resolved at build time? It just a string replacement, like a macro.

cristiand391:
@stomita done here: https://github.com/jsforce/jsforce/pull/1507

>Why not make version be resolved at build time? It just a string replacement, like a macro.

@AllanOricil that version is exposed in the `jsforce` object, we can't resolve it from the package.json at runtime because browser builds don't know about it.

AllanOricil:
@cristiand391 I said "build" time. I was thinking that the string replacement can be done at the end of the build.

cristiand391:
> I said "build" time. I was thinking that the string replacement can be done at the end of the build.

@AllanOricil ah, sorry I misread that. We want this to always match what's on package.json on published builds but it's not necessary in local (we shouldn't publish npm from local unless necessary).

AllanOricil:
@cristiand391 

My suggestion was to set that src/VERSION.ts value at build time with the value of package.json. During a build, if the string was not replaced, the build would throw an exception, which would also cause the release and publish not to happen.

format -> lint -> test -> build -> release -> publish

cristiand391:
closing in favor of #1507 (already merged)

@stomita all new releases will automatically bump the version file:
https://github.com/jsforce/jsforce/commit/3c9e4e6c658a3660c5038c53745283215313138a

PR #1507, opened 2024-06-11
fix: `jsforce.VERSION` matches current version

cristiand391:
Fixes: https://github.com/jsforce/jsforce/issues/1498

This PR adds a pre-commit hook to bump the package version in `src/VERSION.ts` to the one that will be published in the `onRelease` workflow.

### DIFF
--- a/scripts/bump-version-file.js
+++ b/scripts/bump-version-file.js
@@ -1,0 +1,9 @@
+const fs = require('node:fs/promises');
+
+exports.preCommit = async (props) => {
+  const updatedVersion = (await fs.readFile('./src/VERSION.ts', {
+    encoding: 'utf-8'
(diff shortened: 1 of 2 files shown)


scripts/bump-version-file.js
```diff
@@ -0,0 +1,9 @@
+const fs = require('node:fs/promises');
+
+exports.preCommit = async (props) => {
+  const updatedVersion = (await fs.readFile('./src/VERSION.ts', {
+    encoding: 'utf-8'
+  })).replace(/'[^']*'/g, `'${props.version}'`)
```
cristiand391, line 6: `props.version` will be the new version to be published so this will always match the one in `package.json`
https://github.com/TriPSs/conventional-changelog-action/?tab=readme-ov-file#pre-commit-hook

PR #1512, opened 2024-06-13
fix: only refresh token on valid scenarios

cristiand391:


cristiand391:
closing in favor of https://github.com/jsforce/jsforce/pull/1517

Issue #1411, opened 2024-04-21
If logging in via username/password and password has expired SDK enters infinite login loop


AllusiveBox:
So this was an interesting issue we ran into the other day. Our company has a requirement that passwords be cycled every 90 days. One of our integrations uses a username/password combination to login. Someone noted the date the password expired on incorrectly and as a result our password expired before we reset it. 

This resulted in a scenario that allows us to log in to Salesforce (Salesforce support later told us that this was an intended feature and not a bug), but any time we attempt to operate on any resource, the SDK attempts to reauthenticate due to an invalid auth token. This cycle continues until our integration times out or we hit the login limit and Salesforce then restricts our access due to repeated logins.

While I have feelings about Salesforce treating a login using expired username/password via their API as valid, I think that the SDK should be able to handle this scenario and not retry indefinitely.

samschurter:
This looks like it's probably related to #1291. We ran into this last year and determined the loop was a bug in JSForce when it's handling failed login attempts, but at the time there was no real movement on bugfixes. We ended up bypassing the problem entirely by switching to the Salesforce OAuth2 client credential flow with a Connected App

cristiand391:
Yeah, I think I know where this is happening. I'm finishing the migration guides for jsforce v3 and then will triage GH issues.

cristiand391:
This should be fixed in v3, see: https://github.com/jsforce/jsforce/discussions/1529

anurag-roy:
Still facing this issue on the latest version :(

Issue #1308, opened 2023-02-20
passwordExpired is not handled


thib3113:
Hello,

I didn't see an issue about this, so I create one . 

when doing 
```js
const sfconn = new jsforce.Connection({
    loginUrl: 'https://test.salesforce.com'
  })
await sfconn.login(
  username,
  password`
)
```

It will do a soap call to salesforce, and fail only if status >= 400 . 

My problem was in my case salesforce return 200 with : 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"
    xmlns="urn:partner.soap.sforce.com" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <soapenv:Body>
        <loginResponse>
            <result>
                [...]
                <passwordExpired>true</passwordExpired>
                [...]
            </result>
        </loginResponse>
    </soapenv:Body>
</soapenv:Envelope>
```

and this will success . 

So, if I want to do a query, this will loop trying to do a request, and salesforce telling
```json
[{"message":"Your password has expired. Please reset your password.","errorCode":"INVALID_OPERATION_WITH_EXPIRED_PASSWORD"}]
```

So, I see two problems : 
 - query loop never end (so retrying each times to call salesforce)
 - passwordExpired is not handled on login
 
 
 PS : I know you're working on the V2, and I'm not blocked by this problem, it's just FYI

matthewhenry1:
I've just run into this issue, it's a silent killer.

As shared above, the crux of the issue is that the code base doesn't recognize `passwordExpired` as a response object. Despite being present in the [loginResult](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_login_loginresult.htm).

Since the `login()` method returns a 200, it seems as if the user has authenticated. However until you execute a method like `describe()` with the connection object you'll find that the code base will continuously loop attempting to re-execute the `login()` method despite the password being expired. This will result in a login rate limit being exceeded or worse the account being locked out.

Tediously, the only way to determine if the account has a valid password is to initialize the connection object with the `instanceUrl` and `accessToken` then execute a method like `identity()` to actually verify if the password is valid, only then will you receive: `INVALID_OPERATION_WITH_EXPIRED_PASSWORD: Your password has expired. Please reset your password.`

`
const jsforce = require("jsforce");
const username = "myusername";
const password = "mypassword";
[...]

PR #1517, opened 2024-06-19
fix(login): throw if password expired

cristiand391:
fixes:
* https://github.com/jsforce/jsforce/issues/1308
* https://github.com/jsforce/jsforce/issues/1411

Fix bug where trying to `conn.login` with an expired password would cause an infinite loop.

`conn.login` sets a refresh delegate function:
https://github.com/jsforce/jsforce/blob/2621d0b863487056ce25f949ad8d007125ed04b7/src/connection.ts#L507

the func calls `conn.login` and grabs `conn.accessToken`:
https://github.com/jsforce/jsforce/blob/2621d0b863487056ce25f949ad8d007125ed04b7/src/connection.ts#L185

the API returns 200 with `passwordExpired` set to `true` and a restricted access token (you can use it only to set a new password via `setPassword`):
https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_login_loginresult.htm

so any subsequent request would trigger an infinite loop refreshing the restricted token.

### testing:

1. run `npm test:setup:org` to get a scratch org for testing, you'll get credentials printed
2. run `sf org open --path 'lightning/setup/SecurityExpirePasswords/home' --target-org jsforce-test-org` and expire all passwords

save this in a file at the root of the jsforce dir and run it with your scratch org creds:
```js
const jsforce = require('./lib/index');

const conn = new jsforce.Connection({
  version: '60.0',
  logLevel: 'DEBUG',
  loginUrl: '<instance-url>',
  instanceUrl: '<instance-url>',
});

(async() => {
  try {
    await conn.login('<username>','<password>')

    await conn.describe("Opportunity");

  } catch(err) {
    throw err
  }
})()
```

in the `3.0` branch you should see it enter in the session-refresh loop and with this branch it should fail on the first `conn.login` call.

Added a UT to cover this scenario.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -568,6 +568,17 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
       const faultstring = m && m[1];
       throw new Error(faultstring || response.body);
     }
+    // the API will return 200 and a restriced token when using an expired password:
+    // https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_login_loginresult.htm
+    //
+    // we need to throw here to avoid a possible infinite loop with session refresh where:
+    //  1. login happens, `this.accessToken` is set to the restricted token
+    //  2. requests happen, get back 401
+    //  3. trigger session-refresh (username/password login has a default session refresh delegate function)
+    //  4. gets stuck refreshing a restricted token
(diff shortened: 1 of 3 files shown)


mshanemc (approved):

mshanemc:
approved, merging without QA given the 
1. low risk (limited to a matched conditional) 
2. real-world test coverage 
3. already saw this in demos

PR #1518, opened 2024-06-25
fix(session-refresh): improve error message

cristiand391:
Any failure coming from `refreshFn` should surface as a session-refresh failure.

Example throwing in `refreshFn`:

## Before

Screenshot 2024-06-25 at 5 34 42 PM

## After
Screenshot 2024-06-25 at 5 35 20 PM

### DIFF
--- a/src/session-refresh-delegate.ts
+++ b/src/session-refresh-delegate.ts
@@ -63,6 +63,8 @@ export class SessionRefreshDelegate<S extends Schema> {
       });
       await this._refreshPromise;
       this._logger.info('<refresh complete>');
+    } catch(err) {
+      throw new Error(`Unable to refresh session due to: ${(err as Error).message}`)
     } finally {
       this._refreshPromise = undefined;
       this._lastRefreshedAt = Date.now();
--- a/test/connection-session.test.ts
+++ b/test/connection-session.test.ts
@@ -272,7 +272,7 @@ if (isNodeJS()) {
         assert.fail();
       } catch (err) {
         assert.ok(err instanceof Error);
-        assert.ok(err.name === 'invalid_grant');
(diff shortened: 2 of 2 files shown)


src/session-refresh-delegate.ts
```diff
@@ -63,6 +63,8 @@ export class SessionRefreshDelegate<S extends Schema> {
       });
       await this._refreshPromise;
       this._logger.info('<refresh complete>');
+    } catch(err) {
+      throw new Error(`Unable to refresh session due to: ${(err as Error).message}`)
```
cristiand391, line 67: the error passed in the callback is typed, it's guaranteed to be an instance of `Error`:
https://github.com/jsforce/jsforce/blob/3c9e4e6c658a3660c5038c53745283215313138a/src/types/common.ts#L9

test/connection-session.test.ts
```diff
@@ -272,7 +272,7 @@ if (isNodeJS()) {
         assert.fail();
       } catch (err) {
         assert.ok(err instanceof Error);
-        assert.ok(err.name === 'invalid_grant');
+        assert.ok(err.message === 'Unable to refresh session due to: expired access/refresh token');
```
cristiand391, line 275: without the catch block added this err used to be built from this response:
```
{
  error: 'invalid_grant',
  error_description: 'expired access/refresh token'
}
```

cristiand391:
@WillieRuemmele no need to QA (the modified test ensure the new error msg is thrown), the plugin-org NUTs failure is unrelated, see: https://salesforce-internal.slack.com/archives/G02K6C90RBJ/p1719424602794139

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
✅ : UT passes

PR #1520, opened 2024-06-28
fix: retry 420

mshanemc:
SDR was retrying these for deploy polling, probably should do everywhere
@W-16119908@

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -45,7 +45,7 @@ async function startFetchRequest(
   let retryCount = 0;
 
   const retryOpts: Required<HttpRequestOptions['retry']> = {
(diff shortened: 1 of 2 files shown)


WillieRuemmele (approved):

cristiand391 (approved):

cristiand391:
@mshanemc I used `420` in the retry UTs, you'll need to update those:
https://github.com/jsforce/jsforce/blob/0184e2113962ec9f3193a9a8b41db3ccdfcd44a7/test/http-api.test.ts#L63

`npx jest test/http-api.test.ts` to run them locally, don't need org creds.

cristiand391:
plugin-org NUT failure is unrelated, merging...

PR #1574, opened 2024-09-06
Set correct entrypoint to module in package.json

### LINKED ISSUE #1553: vite:import-analysis Failed to resolve entry for package jsforce
When trying to import jsforce in a vite project i am receiving the error:
`vite:import-analysis Failed to resolve entry for package jsforce`

This appears to be the same as this issue https://github.com/vitejs/vite/issues/7725, which is that the path in the `module` field of the `package.json` file must resolve.

In jsforce:
https://github.com/jsforce/jsforce/blob/c83cbd4b0aa4c1706cbe1edaf1eb73ecf64fdf4d/package.json#L26

This path appears to be invalid? I am not familiar with either the project or the intended use of the `module` path, so I'm not really sure what the right resolution is here?

I can tell you that updating the `module` path to (what appears to me to be) the correct entry point `"module": "./index"` seems to correct the issue.

### LINKED ISSUE #1560: @jsforce/jsforce-node lacks ESM code entry files
Hi!

We realized the library is missing the files for the esm entrypoint:
https://github.com/jsforce/jsforce/blob/506754881681be31daa891d09e8daaa741ceec59/package.json#L26

The folder doesn't exist on github, because it is being gitignored in https://github.com/jsforce/jsforce/blob/506754881681be31daa891d09e8daaa741ceec59/.gitignore#L3

Could you add the missing files and allow esm-first imports?

schwarmco:
Fixes #1560
Fixes #1553

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     }
   ],
   "main": "./index",
-  "module": "./module/index",
+  "module": "./index",
   "browser": {
     ".": "./browser/index.js",
     "./index.js": "./browser/index.js",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     }
   ],
   "main": "./index",
-  "module": "./index",
   "browser": {
     ".": "./browser/index.js",
     "./index.js": "./browser/index.js",


package.json
```diff
@@ -23,7 +23,7 @@
     }
   ],
   "main": "./index",
-  "module": "./module/index",
+  "module": "./index",
```
cristiand391, line 26: we don't ship an ESM build yet so it's better to just remove this field (`module` in pjson make tools think this is the ESM entrypoint, we don't want CJS be treated as ESM).

cristiand391 (changes requested):
hey @schwarmco, sorry for the late reply.

jsforce v3 doesn't ship an ESM build yet so it will be better to just remove the `module` field for now.

schwarmco:
> jsforce v3 doesn't ship an ESM build yet so it will be better to just remove the `module` field for now.

that makes perfect sense. can confirm it still works with the import-syntax.
i've rebased and removed the line from package.json

cristiand391 (approved):

cristiand391:
we got 1 external NUT failure but it's unrelated to this change, merging this...

thanks for the PR @schwarmco :)

PR #1588, opened 2024-10-03
fix(bulk2): `jobComplete` event includes job info

cristiand391:
Makes ingest/query poll `jobComplete` event include job info


iowillhoit (approved):
LG

PR #1603, opened 2024-10-18
feat(bulk2): allow to get raw job results

cristiand391:
Allow to fetch raw CSV results from bulk2 jobs.

The getResults method would fetch the CSV, parse them into JS arrays and cache that in an internal class field, this PR adds support for these methods to return results a CSV (string) and intentionally skip caching to avoid switching types.

Required for:
https://github.com/salesforcecli/plugin-data/pull/1097

@W-12408034@

### DIFF
--- a/src/api/bulk2.ts
+++ b/src/api/bulk2.ts
@@ -863,7 +863,21 @@ export class IngestJobV2<S extends Schema> extends EventEmitter {
    *
    * @returns Promise<IngestJobV2SuccessfulResults>
    */
-  async getSuccessfulResults(): Promise<IngestJobV2SuccessfulResults<S>> {
+  async getSuccessfulResults(raw?: false): Promise<IngestJobV2SuccessfulResults<S>>
+  async getSuccessfulResults(raw: true): Promise<string>
+  async getSuccessfulResults(raw?: boolean): Promise<IngestJobV2SuccessfulResults<S> | string> {
+    const reqOpts: BulkRequest = {
+      method: 'GET',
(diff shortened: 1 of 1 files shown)


src/api/bulk2.ts
```diff
@@ -863,7 +863,21 @@ export class IngestJobV2<S extends Schema> extends EventEmitter {
...
+  async getSuccessfulResults(raw?: boolean): Promise<IngestJobV2SuccessfulResults<S> | string> {
+    const reqOpts: BulkRequest = {
+      method: 'GET',
+      path: `/${this.id}/successfulResults`,
+    }
+
+    if (raw) {
+      return this.createIngestRequest<string>({
+        ...reqOpts,
+        responseType: 'text/plain',
```
cristiand391, line 877: jsforce will check `responseType` ( or fallback to `content-type` header ) to try parse the body request to a JS value, passing `text/plain` so it skip parsing.

src/api/bulk2.ts
```diff
@@ -887,16 +901,29 @@ export class IngestJobV2<S extends Schema> extends EventEmitter {
    *
    * @returns Promise<IngestJobV2SuccessfulResults>
```
WillieRuemmele, line 902: should we update jsdocs to mention this parameter?

WillieRuemmele (approved):

WillieRuemmele:
[QA](https://github.com/salesforcecli/plugin-data/pull/1097#issuecomment-2427710275) ✅

iowillhoit (approved):
LG

PR #1606, opened 2024-10-23
fix(oauth2): don't strip URL path

cristiand391:
Allow the OAuth2 module to handle supported URLs:
> All OAuth endpoints require secure HTTP (HTTPS). For standard OAuth 2.0 authorization flows, these host domains are supported unless otherwise specified.
login.salesforce.com
A My Domain URL, such as https://MyCompany.my.salesforce.com
An Experience Cloud site URL, such as https://MyDomainName.my.site.com
A custom URL
For sandboxes, use test.salesforce.com or the My Domain login URL for the sandbox, such as MyDomainName--SandboxName.sandbox.my.salesforce.com.

https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_endpoints.htm&type=5

### before

A community URL like `https://innovation-momentum-8840-dev-ed.scratch.my.salesforce.com/AppCentral` would get the pathname (`/AppCentral`) trimmed.

### after

URL pathname is kept, jsforce handles possible trialing slash.

@W-17041779@

### DIFF
--- a/src/oauth2.ts
+++ b/src/oauth2.ts
@@ -102,9 +102,12 @@ export class OAuth2 {
     } else {
       const loginUrlObject = new URL(loginUrl || defaultOAuth2Config.loginUrl);
       this.loginUrl = loginUrlObject.href
-      this.authzServiceUrl = `${loginUrlObject.origin}/services/oauth2/authorize`;
-      this.tokenServiceUrl = `${loginUrlObject.origin}/services/oauth2/token`;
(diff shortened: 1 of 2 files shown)


iowillhoit:
This looks pretty good, except if you have a pathname that ends in a `/`. You end up with something like this with two `//`:

https://innovation-momentum-8840-dev-ed.scratch.my.salesforce.com/foobar//services/oauth2/authorize

iowillhoit (approved):

PR #1622, opened 2024-11-14
fix(http): handle more CSV delimiters on parse

cristiand391:
Add known CSV delimiters to the CSV parse helper used to parse a request body.

csv-parse:
https://csv.js.org/parse/options/delimiter/

for `sf data upsert/delete bulk` we call `job.getAllResults()` expecting the CSV result records as JS values but if a user uploaded a CSV file using a delimiter other than `COMMA` then the CSV parser here would fail and we get back the CSV as a string:
https://github.com/jsforce/jsforce/blob/0bbf6313628ff5380138ea5df23b06a59d4fe58c/src/http-api.ts#L220

https://github.com/salesforcecli/plugin-data/pull/1110

@W-16971607@

### DIFF
--- a/test/bulk2.test.ts
+++ b/test/bulk2.test.ts
@@ -48,7 +48,7 @@ function ensureSuccessfulBulkResults(
   assert.ok(Array.isArray(successfulResults));
   assert.ok(successfulResults.length === qty);
 
-  const recordCreated = operation === 'insert' ? 'true' : 'false';
+  const recordCreated = ['insert', 'upsert'].includes(operation) ? 'true' : 'false';
 
   for (const successfulResult of successfulResults) {
     assert.ok(isString(successfulResult.sf__Id));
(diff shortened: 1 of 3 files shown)


WillieRuemmele (dismissed):

test/bulk2.test.ts
```diff
@@ -225,28 +225,50 @@ if (isNodeJS()) {
...
-    const id = Date.now();
+  it('should bulk insert from file and return inserted results', async () => {
+    // insert 100 account records from csv file
+    const csvStream = fs.createReadStream(
+      path.join(__dirname, 'data', 'Account_bulk2_test.csv'),
+    );
 
-    await insertAccounts(id);
+    const res = await conn.bulk2.loadAndWaitForResults({
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
```
WillieRuemmele, line 235: why not `require` at the top of the file?

WillieRuemmele (approved):

test/bulk2.test.ts
cristiand391: e2e test files are loaded in node and browser (via webpack <-> karma setup) so we can't have top-level, node-only code here.
It's safe to do the import here because this test block only runs on node:
https://github.com/jsforce/jsforce/blob/0bbf6313628ff5380138ea5df23b06a59d4fe58c/test/bulk2.test.ts#L196

failure with top-level import:
https://github.com/jsforce/jsforce/actions/runs/11902422107/job/33167424899

WillieRuemmele:
QA'd with results in plugin-data

PR #1623, opened 2024-11-20
fix: deprecated errorCode prop of type SaveError and added statusCode prop

aaron-csetter:
Hello! 👋

I'm an engineer at nCino, and I noticed the `SaveError` type has an incorrect property of `errorCode` that does not match the return value from the SF APIs. [As per the documentation](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_database_error.htm#apex_methods_system_database_error), the `errorCode` property should be `statusCode`. This is consistent with the actual error returned: 

```
{\"success\":false,\"errors\":[{\"statusCode\":\"UNABLE_TO_LOCK_ROW\",\"message\":\"unable to obtain exclusive access to this record\",\"fields\":[]}]}
```

This PR adds the `statusCode` field and marks the `errorCode` field as deprecated to avoid a breaking change.

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -1131,6 +1131,7 @@ export class Query<
         errors: [
           {
             errorCode: r.sf__Error,
+            statusCode: r.sf__Error,
             message: r.sf__Error,
           },
         ],
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "jsforce",
-  "version": "3.6.1",
+  "version": "3.6.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "jsforce",
-      "version": "3.6.1",
+      "version": "3.6.2",
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.23.1",
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -212,7 +212,7 @@ function createUsernamePasswordRefreshFn<S extends Schema>(
 function toSaveResult(err: SaveError): SaveResult {
   return {
     success: false,
-    errors: [err],
+    errors: [{ ...err, errorCode: err.statusCode }],
   };
 }
 
--- a/src/types/common.ts
+++ b/src/types/common.ts
@@ -66,7 +66,9 @@ export type Record = {
 export type SavedRecord = Record & { Id: string };
 
 export type SaveError = {
+  /** @deprecated use {@link SaveError.statusCode} instead */
   errorCode: string;
+  statusCode: string;
   message: string;
   fields?: string[];
 };


src/query.ts
```diff
@@ -1131,6 +1131,7 @@ export class Query<
         errors: [
           {
             errorCode: r.sf__Error,
+            statusCode: r.sf__Error,
```
cristiand391, line 1134: these are bulk the bulk v2 API, I don't see `statusCode` there:
https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/get_job_failed_results.htm?q=__error

cristiand391:
I still see the rest API return the `errorCode` field on some cases so we shouldn't deprecate it (the updates you made affect sobject creation results).

Screenshot 2024-11-20 at 1 46 01 PM

API docs mention `statusCode` tho:
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_create.htm

do you have an example request where it would return `statusCode` and not `errorCode`?

src/query.ts
aaron-csetter: Hey Cristian, thanks for the review!

I made this change in anticipation of the `errorCode` being a deprecated field, but from your other comment, it seems some APIs return this.

I first noticed this when creating a permSetAssignment record using the following command in jsForce:

```typescript
return await org.sobject('PermissionSetAssignment').create(permissionAssignments);
```

^^ this returned the error in my initial PS post above.

To further investigate this, I went ahead and made the same request via the Rest API and it actually returned error code instead of status code 😆 :
```
MALFORMED_ID: PermissionSet ID: id value of incorrect type: 0
    at HttpApi.getError (/Users/aaron.csetter/repos/sandbox/node_modules/@jsforce/jsforce-node/lib/http-api.js:290:56)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/aaron.csetter/repos/sandbox/node_modules/@jsforce/jsforce-node/lib/http-api.js:127:33
    at async main2 (/Users/aaron.csetter/repos/sandbox/dist/index.js:36:17) {
  data: {
    message: 'PermissionSet ID: id value of incorrect type: 0',
    errorCode: 'MALFORMED_ID',
    fields: [ 'PermissionSetId' ]
  },
  errorCode: 'MALFORMED_ID'
}
```

I haven't looked into what underlying API that `Connection.sobject()` is using to create records, but it's likely NOT the rest API based on what you and I are seeing, haha.

aaron-csetter: Hey @cristiand391 

Looks like when multiple records are passed in the the `Connection.create()` method, the request uses [sObject Collections](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_create.htm), which even the documentation says `statusCode` is returned in the error:

PR #1636, opened 2025-01-20
fix(bulk): update JobInfo type

cristiand391:
Updates Bulk v1 JobInfo type with more properties:
https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/asynch_api_reference_jobinfo.htm

All these are of type string due to how we convert XML -> JSON in our HTTP utility, see comments in code.
We can't fix properly parsing numbers now b/c consumers might be accessing them as strings already.

Also fix Chrome perms in the test workflow, I thought it was related to [this PR](https://github.com/jsforce/jsforce/pull/1628) but reverting didn't work. The only difference I saw in that workflow passing and the one in this PR's first commit is that GHA bumped the ubuntu image (20 -> 22). 🤷🏼 

https://github.com/salesforcecli/plugin-data/pull/1172
@W-17606312@

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,6 +18,12 @@ jobs:
       - run: |
           chromedriver --url-base=/wd/hub &
           sudo Xvfb -ac :99 -screen 0 1280x1024x24 > /dev/null 2>&1 & # optional
+
+      - name: Fix Chrome sandbox permissions
+        run: |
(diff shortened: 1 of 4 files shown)


cristiand391:
unrelated plugin-data NUTs failures, tried re-running them and saw different one failing so I'll merge this.

iowillhoit (approved):
LG

PR #1654, opened 2025-02-27
Add missing active attribute to IdentityInfo

fobin:
As per [documentation](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_using_openid.htm&type=5) there is active property also for identityInfo type:

active	Boolean specifying whether the queried user is active

### DIFF
--- a/src/types/common.ts
+++ b/src/types/common.ts
@@ -614,6 +614,7 @@ export type IdentityInfo = {
   };
   user_type: string;
   language: string;
+  active: boolean;  
 };
 
 export type UserInfo = {


cristiand391:
thanks for the PR @fobin!
would you mind adding `active` here also so it' accessible via `conn.userInfo`?
https://github.com/jsforce/jsforce/blob/658d7f24783327e74c8d5e5b2bfa9e542d718acd/src/connection.ts#L1488

fobin:
@cristiand391 Thanks for checking this out.

I'm not sure about adding this to UserInfo. It's used in these places where we don't get the value
https://github.com/jsforce/jsforce/blob/76bc3edc23417575a906958ebd5b29b71e51daef/src/connection.ts#L165
https://github.com/jsforce/jsforce/blob/76bc3edc23417575a906958ebd5b29b71e51daef/src/connection.ts#L488
https://github.com/jsforce/jsforce/blob/76bc3edc23417575a906958ebd5b29b71e51daef/src/connection.ts#L520
https://github.com/jsforce/jsforce/blob/76bc3edc23417575a906958ebd5b29b71e51daef/src/connection.ts#L597

What do you think?

Issue #1676, opened 2025-04-10
Ability to opt-out of templating behavior during `find(...).update(...)` calls


no-stack-dub-sack:
**TLDR;** Is it possible to add an option to opt-out of templating behavior when updating records that are meant to contain templating syntax that jsforce otherwise treats specially?

In rare circumstances, the templating syntax can break deployments / result in bugs. Specifically, when deploying a text field that is meant to contain JavaScript code or other strings meant to retain templating syntax. In my use-case, I'm working with Salesforce CPQ, and unfortunately, the way to extend certain functionality is by writing custom JavaScript plugins that are stored in Long Text Area fields. We have a custom repo setup to manage these plugins, and we use jsforce to deploy the changes to the box. However, we are forced to make two requests (one to find the record's id, and one to update it) instead of using `find().update()`, as `sObject().update()` is exempt from the templating behavior.

Example:

```js
const personName = 'Pete';

 // the field is meant to contain code, but the template is stripped when no matching field is found.
const codeString = 'console.log(`Hello, ${personName}!`)';

const results = await conn.sobject('SBQQ__CustomScript__c')
  .find({ Name: 'foo' })
  .update({ SBQQ__Code__c: codeString });
```

The field gets updated with the following:

```
Hello, 
```
While this use-case is probably rare, the CPQ package is a popular one (based on my understanding), and I'm sure there are other use-cases as well where it would be helpful to opt-out of this. Maybe something like: 

```js
const results = await conn.sobject('SBQQ__CustomScript__c')
  .find({ Name: 'foo' })
  .update({ SBQQ__Code__c: codeString }, { templates: false });
```

cristiand391:
> Is it possible to add an option to opt-out of templating behavior when updating records that are meant to contain templating syntax that jsforce otherwise treats specially?

Yep, I saw it was already supported in the record stream implementation but not exposed, got it here: https://github.com/jsforce/jsforce/pull/1679

no-stack-dub-sack:
Awesome, thanks! What about the docs? If you point me in the right direction, I can create a PR. Would be great to get this option exposed there as well. Thanks again for the quick turnaround!

cristiand391:
> What about the docs? If you point me in the right direction, I can create a PR

Sure, it can be mentioned here:
https://jsforce.github.io/document/#update-delete-queried-records

source for the docs:
https://github.com/jsforce/jsforce.github.io/blob/main/src/partials/document/query.html.md

PR #1679, opened 2025-04-14
feat: allow skipping template eval on query update W-18268091

cristiand391:
Fixes: https://github.com/jsforce/jsforce/issues/1676

When doing `sobject().find().update()`, jsforce evaluates the update fields for possible record templates (jsforce syntax sugar for easy string replacement), for the mentioned use-case in #1676 we now allow to skip it via a new boolean option.

Tests added by Cursor 🪄 (needed some polish but were almost perfect).

@W-18268091@


soridalac (approved):

soridalac:
**QA notes**:
✅ : Test pass on should update records with and without expression evaluation

Issue #1691, opened 2025-05-07
Update `sObject.upsert` to use Composite SObject Collections API when appropriate


crumpstar:
### Summary

Jsforce currently uses the Composite SObject Collections API (available since API version 42) to efficiently batch CRUD operations (e.g., insert, update, delete), but it does **not** do so for `upsert`. This results in unnecessary individual API calls or concurrency issues when upserting medium to large datasets.

According to [Salesforce documentation](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_upsert.htm), the Composite SObject Collections API has supported `upsert` operations since **API version 46**.

### Request

Please enhance `sobject().upsert(records, extIdField)` to use the Composite SObject Collections API the same as other CRUD methods.
This would ensure `upsert` is consistent with how `insert` and `update` are already batched, improving efficiency and avoiding the "Exceeded max limit of concurrent call" or running afoul of api limits.

### Impact

This change would:
- Reduce the number of API calls
- Improve performance for bulk upsert scenarios
- Prevent common concurrency errors when dealing with large record arrays

### Example

```js
// Currently causes many separate calls
conn.sobject('MyObject__c').upsert(largeArray, 'External_Id__c');

// Expected: Chunk into groups of 200 and use composite collection API for upserts

PR #1692, opened 2025-05-07
feat(connection): Update upsert to use sobject collection 

crumpstar:
References jsforce/jsforce#1691

Update upsert to use the sobject collection when appropriate.  
- API Version must be 46 or later
- Honor the `allowRecursive` and `MAX_DML_COUNT`
- The records argument must be an array otherwise fall back to the original method.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1285,6 +1285,54 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
     extIdField: string,
     options: DmlOptions = {},
   ): Promise<SaveResult | SaveResult[]> {
+    return Array.isArray(records)
+    ? // check the version whether SObject collection API is supported (46.0)
+      this._ensureVersion(46)
+      ? this._upsertMany(type, records, extIdField, options)
+      : this._upsertParallel(type, records, extIdField, options)
+    : this._upsertParallel(type, records, extIdField, options);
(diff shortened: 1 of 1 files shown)


src/connection.ts
```diff
@@ -1284,6 +1284,72 @@ export class Connection<S extends Schema = Schema> extends EventEmitter {
...
+      if (!sobjectType) {
+        throw new Error('No SObject Type defined in record');
+      }
+      return { [extIdField]: extId, attributes: { type: sobjectType }, ...rec };
+    });
+    let url =
+      [this._baseUrl(), 'composite', 'sobjects', type, extIdField].join('/');
+    if (options.allOrNone) {
+      url += '&allOrNone=true';
+    }
```
cristiand391, line 1332: nit:
this URL param can be removed, only `DELETE` seems to respect it:
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_delete.htm

create/update/upsert take `allOrNone` in the request body.

cristiand391:
@crumpstar thanks for contributing this feature!

I'l merge it into a separate branch to get our E2E running (disabled on forks), then add more coverage and merge to main.

cristiand391 (approved):

PR #1698, opened 2025-05-21
fix(SOAP): block connections using JWT tokens W-18585436

cristiand391:
Update the SOAP module to throw if the provided `connection` is using a JWT-based access tokens.

Due to how session-refresh works and SOAP APIs not support JWT tokens, trying to use them makes jsforce enter into an infinite loop trying to fetch a valid token.

### QA instructions
1. checkout this PR and build jsforce-node: `npm i && npm run jsforce-node:dev && npm run build:node:cjs`
2. link it into plugin-deploy-retrieve or plugin-org: `yarn link @jsforce/jsforce-node`
3. auth to an org using a connected app with JWT-access tokens enabled.
4. run a command that does a SOAP request like `project deploy start`/`org list metadata`:

without this PR, jsforce enters the infinite loop (set `JSFORCE_LOG_LEVEL=DEBUG` to debug logs).
Now any code trying to do SOAP requests will fail like this:

Screenshot 2025-05-21 at 13 25 16

@W-18585436@

### DIFF
--- a/src/util/jwt.ts
+++ b/src/util/jwt.ts
@@ -1,0 +1,18 @@
+/**
+ * Checks if a given access token is a JWT.
+ *
+ * @param {string} accessToken - The access token to check
+ * @returns {boolean} True if the token is a valid JWT token, false otherwise.
+ */
+export function isJWTToken(accessToken: string): boolean {
+  const parts = accessToken.split('.');
+  if (parts.length !== 3) return false;
+
+  const header = parts[0];
+  try {
+    JSON.parse(atob(header));
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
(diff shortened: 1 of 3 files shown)


src/util/jwt.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param {string} accessToken - The access token to check
+ * @returns {boolean} True if the token is a valid JWT token, false otherwise.
+ */
+export function isJWTToken(accessToken: string): boolean {
+  const parts = accessToken.split('.');
+  if (parts.length !== 3) return false;
+
+  const header = parts[0];
+  try {
+    JSON.parse(atob(header));
```
cristiand391, line 13: JWTs have 3 base64-encoded parts (header,payload and signature) separated by a dot.

if we got here, it could also be an opaque token that has 3 dots in it so we do a final check by decoding the base64 header -> try to `JSON.parse` (if the parse didn't fail then we assume it's a JWT header in JSON and return true)

shetzel (dismissed):

shetzel (approved):

PR #1760, opened 2025-09-08
fix: increase timeout- W-19332431

soridalac:
bump default timeout from 5 min to 10min.

follow up PR: https://github.com/jsforce/jsforce/pull/1745

@W-19332431@


src/request.ts
```diff
@@ -179,16 +179,21 @@ async function startFetchRequest(
...
   try {
     res = await executeWithTimeout(fetchWithRetries, fetchTimeout, () =>
       controller.abort(),
     );
   } catch (err) {
+    if (err instanceof AbortError) {
+      emitter.emit('error', {
+        message: 'Request was aborted due to timeout of 10 minutes.',
+      })
+    }
```
cristiand391, line 196: emitting events is will not surface in CLI.
Could we instead append this msg to the one in the error? like `err.message += 'Request was...'`

cristiand391 (approved):

PR #1772, opened 2025-11-12
fix(http): limit retries on 420 responses W-19971144

cristiand391:
~DEPENDS ON: https://github.com/jsforce/jsforce/pull/1770
After #1770 is merged, switch target branch to main (it points to 1770 now to get tests running)~ DONE

This PR updates the http retry logic to limit retries on 420 responses to a max of **2**.

### Context:
Deleted/expired scratch orgs return 420 and causes a long delay on all requests due to the retry with exponential backoff (max 5 retries by default):

Screenshot 2025-11-12 at 12 49 43

We can't stop retrying on 420 tho, from our experience in the CLI we've seen the Metadata API sometimes return 420 on a deploy/retrieve status check request, so for now we'll just limit to 2 retries (which should happen pretty fast).

#### Could this impact 420 retries during deploy/retrieve polling?
No, SDR has an additional retry list of errors that includes 420 responses coming from jsforce:
https://github.com/forcedotcom/source-deploy-retrieve/blob/cec7757bfb7e302e6f85e6fcef34374175520a87/src/client/metadataTransfer.ts#L260

The current behavior when retrying 420 during poll checks works like this:

1. SDR issues deploy/retrieve checkStatus calls -> req done via jsforce
2. jsforce gets 420 res, retries 5 times with exponential backoff -> returns success or throws `ERROR_HTTP_420` if the 6th req still fails
3. SDR catches `ERROR_HTTP_420` and issues a retry (as of today there's a max limit of 1000 retries on consecutive errors)

With the changes in this PR we would have step 2 fail quicker but SDR would still retry on step 3.

### Before
requests to deleted/expired orgs would do 5 retries with exponential backoff, taking a few seconds to return the real error:
```
➜  jsforce git:(cd/limit-420-retry) time JSFORCE_LOG_LEVEL=DEBUG sf data query -q 'select count() from apexclass'
DEBUG   [query]  config is soql: select count() from apexclass
DEBUG   [query]  >>> Query start >>>
DEBUG   [query]  execute with options {
  headers: {},
  responseTarget: 'QueryResult',
  autoFetch: true,
  maxFetch: 50000,
  scanAll: false
}
DEBUG   [query]  SOQL = select count() from apexclass
DEBUG   [http-api]  <request> method=GET, url=https://customer-platform-9304-dev-ed.scratch.my.salesforce.com/services/data/v65.0/query?q=select%20count()%20from%20apexclass
DEBUG   [fetch]  retrying for the 1 time
DEBUG   [fetch]  reason: statusCode match
DEBUG   [fetch]  retrying for the 2 time
DEBUG   [fetch]  reason: statusCode match
DEBUG   [fetch]  retrying for the 3 time
DEBUG   [fetch]  reason: statusCode match
DEBUG   [fetch]  retrying for the 4 time
DEBUG   [fetch]  reason: statusCode match
DEBUG   [fetch]  retrying for the 5 time
DEBUG   [fetch]  reason: statusCode match
DEBUG   [http-api]  elapsed time: 17940 msec
DEBUG   [http-api]  <response> status=420, url=https://customer-platform-9304-dev-ed.scratch.my.salesforce.com/services/data/v65.0/query?q=select%20count()%20from%20apexclass
DEBUG   [http-api]  html response.body: <!DOCTYPE HTML>
<html lang=en-US>
  errorCode: 'ERROR_HTTP_420'
  ... big html response
}
Querying Data... done
Error (1): HTTP response contains html content.
Check that the org exists and can be reached.
See `error.data` for the full html response.


JSFORCE_LOG_LEVEL=DEBUG sf data query -q 'select count() from apexclass'  0.74s user 0.14s system 4% cpu 19.259 total
```

### After

only retry 2 times on 420, then throw:
Screenshot 2026-01-13 at 14 04 20

### Benchmark

`./bin/run.js` with this branch of jsforce-node
sf version: @salesforce/cli/2.118.20

Screenshot 2026-01-13 at 14 02 05

### Testing instructions

1. create a scratch org with duration 1 day and let it expire
alternative:
delete the `ScratchOrgInfo` record from the hub:
```
sf org display -o <active-scratch-org>
# grab org ID and find the activescratch record

sf data query --query "SELECT id FROM ActiveScratchOrg WHERE scratchorg='<org-id>'" -o <hub>

# delete scratch record, ID from previous command starts with 2As
sf data delete record --sobject ActiveScratchOrg --record-id 2AS*************** -o <hub> 
```

5. verify the long delay:
```
JSFORCE_LOG_LEVEL=DEBUG sf data query -q 'select count() from apexclass' -o <expired-scratch>
```

6. checkout this PR and switch to the jsforce-node pkg: `npm run jsforce-node:dev`
7. build: `npm run build:node:cjs`
8. `yarn link` it into plugin-data, then link plugin data
9. repeat step 2, should see only 2 retries

Ref:
https://github.com/forcedotcom/cli/issues/3425
@W-19971144@

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -68,14 +68,26 @@ async function startFetchRequest(
     ],
   };
 
-  const shouldRetryRequest = (
+  const shouldRetryRequest = async (
     maxRetry: number,
     resOrErr: Response | Error | FetchError,
-  ): boolean => {
+  ): Promise<boolean> => {
     if (!retryOpts.methods.includes(request.method)) return false;
 
     if (resOrErr instanceof Response) {
-      if (retryOpts.statusCodes.includes(resOrErr.status)) {
+      // REST API status codes: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/errorcodes.htm
+      //
+      // Deleted/expired scratch orgs return 420 and causes a long delay on all requests due to the retry with exponential backoff.
+      // We still want to retry on 420 (Medatata API requests sometimes would return it) so here we'll limit to a maximum of 2 retries.
+      if (resOrErr.status === 420 && resOrErr.headers.get('content-type') === 'text/html') {
+        const salesforceEdgeRes = (await resOrErr.text()).includes('If it takes too long, please contact support or visit our');
+
+        if (salesforceEdgeRes && retryOpts.maxRetries > 0 && retryCount == 2) {
+          return false
+        } else {
+          return true
+        }
+      } else if (retryOpts.statusCodes.includes(resOrErr.status)) {
         if (maxRetry === retryCount) {
           return false
         } else {
@@ -121,7 +133,7 @@ async function startFetchRequest(
 
     try {
       const res = await fetch(url, fetchOpts);
-      if (shouldRetryRequest(retryOpts.maxRetries, res)) {
+      if (await shouldRetryRequest(retryOpts.maxRetries, res)) {
         logger.debug(`retrying for the ${retryCount + 1} time`);
         logger.debug('reason: statusCode match');
 
@@ -149,7 +161,7 @@ async function startFetchRequest(
         throw error;
       }
 
-      if (shouldRetryRequest(retryOpts.maxRetries, error)) {
+      if (await shouldRetryRequest(retryOpts.maxRetries, error)) {
         logger.debug(`retrying for the ${retryCount + 1} time`);
         logger.debug(`Error: ${(err as Error).message}`);
 
--- a/src/request.ts
+++ b/src/request.ts
@@ -69,10 +69,10 @@ async function startFetchRequest(
     ],
   };
 
-  const shouldRetryRequest = async (
+  const shouldRetryRequest = (
     maxRetry: number,
     resOrErr: Response | Error | FetchError,
-  ): Promise<boolean> => {
+  ): boolean => {
     if (!retryOpts.methods.includes(request.method)) return false;
 
     if (resOrErr instanceof Response) {
@@ -128,7 +128,7 @@ async function startFetchRequest(
 
     try {
       const res = await fetch(url, fetchOpts);
-      if (await shouldRetryRequest(retryOpts.maxRetries, res)) {
+      if (shouldRetryRequest(retryOpts.maxRetries, res)) {
         logger.debug(`retrying for the ${retryCount + 1} time`);
         logger.debug('reason: statusCode match');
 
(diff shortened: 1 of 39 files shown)


src/request.ts
```diff
@@ -68,14 +68,26 @@ async function startFetchRequest(
...
+      //
+      // Deleted/expired scratch orgs return 420 and causes a long delay on all requests due to the retry with exponential backoff.
+      // We still want to retry on 420 (Medatata API requests sometimes would return it) so here we'll limit to a maximum of 2 retries.
+      if (resOrErr.status === 420 && resOrErr.headers.get('content-type') === 'text/html') {
+        const salesforceEdgeRes = (await resOrErr.text()).includes('If it takes too long, please contact support or visit our');
+
+        if (salesforceEdgeRes && retryOpts.maxRetries > 0 && retryCount == 2) {
+          return false
+        } else {
+          return true
```
iowillhoit: This will get into a retry loop here if the `salesforceEdgeRes` text from the server ever changes.
- The status will be `420`
- The content type will be `text/html`
  - Makes it into `if` on `82`
- `salesforceEdgeRes` would be `false` if the server response changes
  - then `shouldRetryRequest` would always return `true`

iowillhoit (changes requested):
See comments

src/request.ts
```diff
@@ -75,7 +76,13 @@ async function startFetchRequest(
     if (!retryOpts.methods.includes(request.method)) return false;
 
     if (resOrErr instanceof Response) {
-      if (retryOpts.statusCodes.includes(resOrErr.status)) {
+      // REST API status codes: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/errorcodes.htm
+      //
+      // Deleted/expired scratch orgs return 420 and causes a long delay on all requests due to the retry with exponential backoff.
+      // We still want to retry on 420 (Metadata API requests sometimes return it) so here we'll limit to a maximum of 2 retries.
+      if (resOrErr.status === 420) {
+        return retry420Count < 2;
```
cristiand391: the firsts commit had a logic bug were I was assuming `retryCount` would count only for 420 responses, but we could still have something lke:

`400 -> 540 -> 420` where `retryCount` would be 2 when the first `420` happened

so now we have a separate counter for it (`retryCount` still counts all retries)

src/request.ts
```diff
@@ -168,12 +178,7 @@ async function startFetchRequest(
       }
 
       logger.debug('Skipping retry...');
-
-      if (maxRetry === retryCount) {
-        throw err;
-      } else {
-        throw err;
-      }
+      throw err;
```
cristiand391: not sure why I added this if/else before 🤷🏼

soridalac (approved):

PR #1784, opened 2026-01-15
fix(http): reduce buffer allocations when reading responses W-18649787

cristiand391:
This PR updates the `MemoryWriteStream` writable that's used to process an HTTP response body.

### Context

We got a report from a customer where a retrieve of static resources via `sf` was constantly hanging in the middle of the polling checks, last log lines printed:
```
DEBUG   [http-api]  missing 'content-length' header, setting it to: 708
DEBUG   [http-api]  <request> method=POST, url=https://dreamhouse-lala.salesforce.com/services/Soap/m/63.0
DEBUG   [http-api]  elapsed time: 75 msec
DEBUG   [http-api]  <response> status=200, url=https://dreamhouse-lala.salesforce.com/services/Soap/m/63.0
DEBUG   [http-api]  missing 'content-length' header, setting it to: 708
DEBUG   [http-api]  <request> method=POST, url=https://dreamhouse-lala.salesforce.com/services/Soap/m/63.0
DEBUG   [http-api]  elapsed time: 1358512 msec
DEBUG   [http-api]  <response> status=200, url=https://dreamhouse-lala.salesforce.com/services/Soap/m/63.0
```

which shows the request promise in L107 was never resolved:
https://github.com/jsforce/jsforce/blob/6c6cc289950a878b8f5f2a81160ad23ea48b7190/src/http-api.ts#L94-L107

**IMPORTANT:**
At this point the CLI didn't thrown an error or exited, the process would just hang indefinitely.

The retrieve polling involves calling `checkRetrieveStatus` multiple times with until it's done, with the last response including the base64-encoded zip of metadata.
 

### Bug
The behavior of `checkRetrieveStatus` and the logs showed that last response with the zip file was causing jsforce to hang (the customer who reported this was on a low-resources windows vm).

code path of the API call:
```
checkRetrieveStatus()
    → _invoke()
    → SOAP.invoke()
    → HttpApi.request()
    → Transport.httpRequest()
    → node-fetch
    → res.body.pipe(output)
    → readAll() with MemoryWriteStream
    → parseXML() with xml2js
```

After debugging it with the payload we found the issue was in the way the `MemoryWriteStream` writer was processing the chunks, calling `Buffer.concat` 1 time per chunk to store the whole response in a buffer was constantly copying/destroying buffers which caused node's garbage collector to froze the process trying to clean up memory.

We verified the fix with the customer and got the retrieve working in ~3minutes.

### Testing
There's a new perf test to measure `readAll` performance, see the how it failed in the 1st commit (only added the test without the writer fixes).

Screenshot 2026-01-15 at 16 30 11

@W-18649787@

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,9 +114,13 @@ jobs:
 
       - name: Install dependencies
         run: npm install
+      
+      - name: Run Perf tests
+        run: npm run test:node:perf
 
       - name: Install sf CLI
-        run: npm install --global @salesforce/cli
+        id: install-sf-cli
+        run: npm install --global @salesforce/cli@nightly
 
       - name: Setup scratch org
         env:
@@ -139,7 +143,7 @@ jobs:
           npm run test:node
 
       - name: Delete scratch org
-        if: always()
+        if: always() && steps.install-sf-cli.outcome == 'success'
         run: sf org delete scratch --target-org jsforce-test-org --no-prompt
 
   salesforce-cli-external-nuts:
@@ -184,7 +188,11 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      - name: Run Perf tests
+        run: npm run test:node:perf
+
       - name: Install sf CLI
+        id: install-sf-cli
         run: npm install --global @salesforce/cli@nightly
 
       - name: Setup scratch org
@@ -208,5 +216,5 @@ jobs:
           npm run test:node
 
       - name: Delete scratch org
-        if: always()
+        if: always() && steps.install-sf-cli.outcome == 'success'
         run: sf org delete scratch --target-org jsforce-test-org --no-prompt
--- a/src/util/stream.ts
+++ b/src/util/stream.ts
@@ -1,4 +1,7 @@
 import { Duplex, PassThrough, Readable, Writable } from 'stream';
+import { getLogger } from './logger';
+
+const logger = getLogger('stream');
 
 export function createLazyStream() {
   const ins = new PassThrough();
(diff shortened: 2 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -114,9 +114,13 @@ jobs:
...
       - name: Install dependencies
         run: npm install
+      
+      - name: Run Perf tests
+        run: npm run test:node:perf
 
       - name: Install sf CLI
-        run: npm install --global @salesforce/cli
+        id: install-sf-cli
+        run: npm install --global @salesforce/cli@nightly
```
cristiand391: the windows job is already installing nightly, we do the same in NUTs.

src/util/stream.ts
```diff
@@ -16,40 +19,55 @@ export function createLazyStream() {
...
+    for (const { chunk } of data) {
+      this._chunks.push(chunk);
+      this._totalBytes += chunk.length;
+    }
     callback();
   }
 
   toString(encoding: BufferEncoding = 'utf-8') {
-    return this._buf.toString(encoding);
+    return Buffer.concat(this._chunks).toString(encoding);
```
cristiand391: this method only gets called when the response is already read:
https://github.com/jsforce/jsforce/blob/6c6cc289950a878b8f5f2a81160ad23ea48b7190/src/util/stream.ts#L52

test/perf/stream-performance.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  // Maximum scaling factor when data size increases 4x (5MB → 20MB)
+  // O(n) should be ~4x, O(n²) would be ~16x
+  // We use 8x as threshold to catch quadratic regressions
+  maxScalingFactor: 8,
+};
+
+/**
+ * Creates a readable stream that delivers data in chunks (simulates network).
+ */
+function createChunkedStream(
```
cristiand391: so at the start of the investigation I was testing this by using nock to return a big response but turns out that counts as 1 chunk!
so we cannot rely on nock for testing perf of this logic around streaming.

shetzel (approved):

PR #1798, opened 2026-06-01
fix(browser): always use fetch in browser/web workers

jonnyhork:
## Why

`src/browser/request.ts` currently picks the transport with:

```ts
if (typeof window !== 'undefined' && typeof window.fetch === 'function') {
  startFetchRequest(...);
} else {
  startXmlHttpRequest(...);
}
```

That gate looks at `window`, but for VS Code Extensions using JSForce`window` is `undefined` inside a Web Worker realm — so jsforce silently falls through to `XMLHttpRequest` whenever it runs in a worker. That's surprising in two ways:

- Worker realms have a working global `fetch`. Every browser jsforce supports today exposes `fetch` on both the window and worker globals, so the XHR path is dead weight in the worker case.
- Some embeddings (e.g. VS Code's web extension host worker) shim `fetch` but not `XMLHttpRequest`, which means the XHR fallback throws on construction. The result is that perfectly-modern environments fail with a transport error that looks like a jsforce bug rather than a missing primitive.

## What

- Drop the `typeof window` gate and always call `startFetchRequest`.
- Remove the now-unused `startXmlHttpRequest` and `getResponseHeaderNames` helpers (file-local; no other references in `src/`).

The fetch path was already the preferred branch in browsers; this change just extends it to worker realms and drops the dead XHR fallback. No public API change.

## Validation

Validated by running VS Code extensions that depend on jsforce against a `patch-package` build of jsforce with this exact change applied, inside a VS Code web extension host worker. Salesforce REST API calls behave identically to the previous XHR path — same URLs, same auth headers, same response shapes — and the worker realm no longer falls through to a missing `XMLHttpRequest` constructor.

jsforce fetch

Local checks on this PR's branch:

- `npm run build` — green
- `npm run lint` — green
- `npm run typecheck` — green
- `npm run test:node` — non-integration suites pass (8/24 suites, 70 tests). The remaining 16 suites fail with `URL_NOT_RESET: Destination URL not reset` because they require a live Salesforce org, which I don't have set up locally. Happy to re-run against a sandbox if you'd like — just let me know how you'd like that wired.

## Compatibility

`fetch` has been baseline in every evergreen browser since 2017 and is exposed on both `Window` and `WorkerGlobalScope`, so this should be a no-op for window-realm consumers and a strict improvement for worker-realm consumers.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,7 +113,10 @@ jobs:
         run: npm install
 
       - name: Install Puppeteer Chrome
-        run: npx puppeteer browsers install chrome
+        shell: bash
+        run: |
+          rm -rf "$HOME/.cache/puppeteer/chrome"
+          npx puppeteer browsers install chrome
 
       - name: Run Perf tests
         run: npm run test:node:perf
@@ -189,7 +192,10 @@ jobs:
         run: npm install
 
       - name: Install Puppeteer Chrome
-        run: npx puppeteer browsers install chrome
+        shell: bash
+        run: |
+          rm -rf "$HOME/.cache/puppeteer/chrome"
+          npx puppeteer browsers install chrome
 
       - name: Run Perf tests
         run: npm run test:node:perf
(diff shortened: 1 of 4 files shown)


mshanemc (dismissed):
Image

nice ratio!

cristiand391 (approved):

PR #1807, opened 2026-06-22
fix: switched from node-fetch to native fetch API @W-23121570@

### LINKED ISSUE #1806: Node.js 24.17.0: jsforce request promise can hang with keep-alive transport
jsforce requests can hang indefinitely on Node.js `24.17.0`.

The first visible failure in our app was:

```js
await conn.authorize({
  grant_type: 'client_credentials',
});
```

After isolating the issue, OAuth auth and a normal REST query both work when jsforce sends `connection: close`.

The same code works on Node.js `24.16.0`.

## Environment

- jsforce versions inspected/tested:
  - `3.6.6`
  - `3.7.0`
  - `3.9.1`
  - `3.10.16`
- `node-fetch`: `2.7.0`
- Node.js:
  - fails: `24.17.0`
  - works: `24.16.0`
- Platform:
  - amd64
  - arm64

## Reproduction

```js
import jsforce from 'jsforce';

const conn = new jsforce.Connection();

conn.oauth2 = new jsforce.OAuth2({
  clientId: process.env.SALESFORCE_CLIENT_ID,
  clientSecret: process.env.SALESFORCE_CLIENT_SECRET,
  loginUrl: 'https://<my-domain>.develop.my.salesforce.com',
});

await conn.authorize({
  grant_type: 'client_credentials',
});

const result = await conn.query(
  'SELECT Id, CaseNumber, Status, Subject, LastModifiedDate FROM Case LIMIT 1',
);

console.log(result.records[0]);
```

## Observed Behavior

On Node.js `24.17.0`, the jsforce request promise can remain unsettled.

In our validation:

- unpatched jsforce on Node.js `24.17.0` left `conn.authorize({ grant_type: 'client_credentials' })` unsettled.
- adding only `content-length` to the OAuth token POST did not fix the issue.
- adding `connection: close` fixed OAuth auth.
- adding `connection: close` in the shared jsforce request transport fixed both OAuth auth and a normal Case query.

## Expected Behavior

jsforce requests should either complete successfully or fail with a bounded request error. They should not hang indefinitely on Node.js `24.17.0`.

## Suspected Cause

jsforce uses `node-fetch@2` for its Node request transport. Node.js `24.17.0` uses keep-alive by default for `http.globalAgent` / `https.globalAgent` and includes HTTP/TLS security changes, including an `http.Agent` idle socket data guard and an `llhttp` update.

This appears to expose an incompatibility between jsforce's current request transport and Node.js `24.17.0` connection reuse behavior.

## Diagnostic Workaround

In `lib/request.js`, setting `connection: close` for Node requests avoids the hang:

```js
const headers = {
  ...(rrequest.headers ?? {}),
};

if (
  !('window' in globalThis) &&
  !('self' in globalThis) &&
  !Object.keys(headers).some((headerName) => headerName.toLowerCase() === 'connection')
) {
  headers.connection = 'close';
}

const fetchOpts = {
  ...rrequest,
  headers,
  ...(input && /^(post|put|patch)$/i.test(request.method)
    ? { body: input }
    : {}),
  redirect: 'manual',
  signal: controller.signal,
  agent,
};
```

Local validation showed that `content-length` alone does not fix the issue. `connection: close` is the only local workaround we found that avoids the hang, but it is not intended as the long-term jsforce fix.

## Validation Result

With `connection: close` applied in the shared request transport, Node.js `24.17.0` successfully completed:

- OAuth `client_credentials` auth
- `SELECT Id, CaseNumber, Status, Subject, LastModifiedDate FROM Case LIMIT 1`

The query returned one Case record successfully.

## Related Node.js Changes

Node.js `24.17.0` release:
https://nodejs.org/en/blog/release/v24.17.0

Relevant commits:

- `http.Agent` response queue poisoning fix:
  https://github.com/nodejs/node/commit/be7e719c3f85e7a41a5d2fbf5bf094556e2ccd02
- `llhttp` update to `9.4.2`:
  https://github.com/nodejs/node/commit/cb2aed980c4791f3aef79f025851c21ac78c360a

jfeingold35:
Should close #1806 and https://github.com/forcedotcom/cli/issues/3586

### DIFF
--- a/.github/workflows/externalNut.yml
+++ b/.github/workflows/externalNut.yml
@@ -58,7 +58,7 @@ jobs:
         name: cli install
         with:
           max_attempts: ${{ inputs.attempts }}
-          command: npm install -g @salesforce/cli@dev --omit=dev
+          command: npm install -g @salesforce/cli@nightly --omit=dev
           timeout_minutes: 20
       - uses: nick-fields/retry@943e742917ac94714d2f408a0e8320f2d1fcafcd
         name: git clone project
--- a/.github/workflows/externalNut.yml
+++ b/.github/workflows/externalNut.yml
@@ -58,7 +58,7 @@ jobs:
         name: cli install
         with:
           max_attempts: ${{ inputs.attempts }}
-          command: npm install -g @salesforce/cli@nightly --omit=dev
+          command: npm install -g @salesforce/cli@dev --omit=dev
           timeout_minutes: 20
       - uses: nick-fields/retry@943e742917ac94714d2f408a0e8320f2d1fcafcd
         name: git clone project
--- a/package.json
+++ b/package.json
@@ -249,7 +249,7 @@
     "@types/inquirer": "^6.5.0",
     "@types/jest": "^29.5.5",
     "@types/multistream": "^2.1.1",
-    "@types/node": "^18.15.3",
+    "@types/node": "^22.20.0",
     "@types/shelljs": "^0.8.15",
     "@types/through2": "^2.0.34",
     "@types/xml2js": "^0.4.5",
@@ -297,10 +297,5 @@
     "webpack-cli": "^5.0.2",
     "wireit": "^0.14.12",
     "zx": "^7.2.3"
-  },
-  "overrides": {
-    "node-fetch": {
-      "whatwg-url": "^14.0.0"
-    }
   }
 }
--- a/package.json
+++ b/package.json
@@ -233,7 +233,6 @@
     "csv-stringify": "^6.6.0",
     "faye": "^1.4.0",
     "form-data": "^4.0.4",
-    "https-proxy-agent": "^5.0.0",
     "inquirer": "^8.2.7",
     "multistream": "^3.1.0",
     "open": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "database.com"
   ],
   "homepage": "http://github.com/jsforce/jsforce",
-  "version": "3.10.17-dev.0",
+  "version": "3.10.16",
   "repository": {
     "type": "git",
     "url": "git://github.com/jsforce/jsforce.git"
@@ -216,7 +216,7 @@
     "index.d.ts"
   ],
   "engines": {
-    "node": ">=18"
+    "node": ">=22"
   },
   "bin": {
     "jsforce": "./bin/jsforce",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "database.com"
   ],
   "homepage": "http://github.com/jsforce/jsforce",
-  "version": "3.10.16",
+  "version": "3.10.16-dev.0",
   "repository": {
     "type": "git",
     "url": "git://github.com/jsforce/jsforce.git"
--- a/package.json
+++ b/package.json
@@ -236,8 +236,8 @@
     "https-proxy-agent": "^5.0.0",
     "inquirer": "^8.2.7",
     "multistream": "^3.1.0",
-    "node-fetch": "^2.6.1",
     "open": "^7.0.0",
+    "undici": "^8.5.0",
     "xml2js": "^0.6.2"
   },
   "devDependencies": {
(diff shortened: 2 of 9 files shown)


nbolton:
Nice. For anyone that needs a workaround right now before this lands...

```ts
import jsforce from "jsforce";

/**
 * Workaround for jsforce#1806. Node 24.17.0 made keep-alive the default for the global HTTP
 * agents, and jsforce's bundled (unmaintained) node-fetch@2 hangs forever reusing those
 * sockets, so every Salesforce request times out. Forcing `Connection: close` gives each
 * request a fresh socket, which sidesteps it. Delete this file once jsforce#1807 ships and we
 * bump jsforce.
 */

type Connection = InstanceType<typeof jsforce.Connection>;

export async function createSalesforceConnection(
  username: string,
  password: string,
): Promise<Connection> {
  const connection = new jsforce.Connection({});

  const transport = connection._transport;
  const originalHttpRequest = transport.httpRequest.bind(transport);
  transport.httpRequest = (req, options) => {
    req.headers = { ...req.headers, connection: "close" };
    return originalHttpRequest(req, options);
  };

  await connection.login(username, password);
  return connection;
}
```

nbolton (approved):

src/request.ts
```diff
@@ -186,18 +184,21 @@ async function startFetchRequest(
...
 
   try {
     res = await executeWithTimeout(fetchWithRetries, fetchTimeout, () =>
       controller.abort(),
     );
   } catch (err) {
-    if (err instanceof AbortError) {
-      (err as Error).message += ' Request was aborted due to timeout of 10 minutes.';
+    let throwableError: Error;
+    if (err instanceof DOMException && err.name === 'AbortError') {
```
jfeingold35, line 196: `DOMException`'s `message` property is not editable, so we just need to construct a new one in this case.

.github/workflows/externalNut.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         name: cli install
         with:
           max_attempts: ${{ inputs.attempts }}
-          command: npm install -g @salesforce/cli@nightly --omit=dev
+          command: npm install -g @salesforce/cli@dev --omit=dev
```
WillieRuemmele: revert these?

package.json
```diff
@@ -236,8 +236,8 @@
     "https-proxy-agent": "^5.0.0",
```
WillieRuemmele: I'm not sure if we're using `https-proxy-agent` anymore now

src/request.ts
```diff
@@ -186,18 +184,21 @@ async function startFetchRequest(
...
+      throwableError = new DOMException((err as Error).message + ' Request was aborted due to timeout of 10 minutes.', (err as Error).name);
```
WillieRuemmele: my claude is saying default timeout is 30 mins?

test/http-api.test.ts
```diff
@@ -40,13 +55,11 @@ describe('HTTP API', () => {
...
     it('returns response after retry limit is reached', async () => {
-      nock(loginUrl)
-        .get('/services/data/v59.0')
-        .times(4)
-        .reply(429, JSON.stringify({
-          errorCode: 'INTERNAL_SERVER_ERROR',
-          message: 'Invalid AiEvaluation identifier'
-        }))
+      const pool = mockAgent.get(loginUrl);
+      pool.intercept({ path: '/services/data/v59.0', method: 'GET' }).reply(429, JSON.stringify({
```
WillieRuemmele, line 59: my claude is saying: 
```
  - Test error mocks may not match undici's error wrapping — test/http-api.test.ts:104,116,135 — The retry logic expects
  TypeError → cause → SocketError → cause → {code}, but replyWithError(Object.assign(new Error('ECONNRESET'), {code: 
  'ECONNRESET'})) produces a flat error. If tests pass, it's because undici's MockAgent wraps these into TypeErrors, but
  this is fragile and implicit.
```

WillieRuemmele (changes requested):

package.json
jfeingold35: Good catch.

test/http-api.test.ts
jfeingold35: I tried wrapping it in a TypeError and that didn't work, so I think it might actually be correct for it to be unwrapped here. Regardless, I kind of don't want to mess with something that's currently working right now, it that makes any sense.

package.json
WillieRuemmele: one more, I don't think we're using `nock` anymore either, could be worth it to remove and regen package-lock.json after too

jfeingold35: Oh, good catch. I'll pull that one too.

WillieRuemmele (dismissed):

nbolton:
Thanks for fixing this, guys!

PR #1816, opened 2026-07-30
PIM-2093 | Republish as @propelsoftware/jsforce-propel

chandrasub:
## Summary
- Renames the package from `jsforce` to `@propelsoftware/jsforce-propel` and bumps version to 1.9.4 (previously published npm package was at 1.9.3, but that bump was never committed to this repo).
- `jsforce-propel` on npm is currently maintained solely by a departed employee's personal npm account — no other collaborators, same root cause as PIM-2093 for `propel-sfdc-connect`.
- Verified `lib/connection.js` is byte-identical between this repo and the published 1.9.3 tarball, so this rename does not lose any code changes that only existed in the published package.

## Before merging / publishing
- [ ] Merge this PR
- [ ] `npm publish --access public` (matches this repo's existing public visibility) from an `@propelsoftware` org owner account
- [ ] Update consumers (`pim-data-service`, `CloudFileStorage`, possibly others) to depend on `@propelsoftware/jsforce-propel@1.9.4`

## Test plan
- [ ] `npm install` resolves cleanly under the new name
- [ ] Existing test suite still passes (no behavior change, rename + version bump only)

Ref: PIM-2093

### DIFF
--- a/lib/VERSION.js
+++ b/lib/VERSION.js
@@ -1,2 +1,2 @@
 'use strict';
-module.exports = '1.9.1';
+module.exports = '1.9.4';
--- a/lib/api/streaming-extension.js
+++ b/lib/api/streaming-extension.js
@@ -118,7 +118,7 @@ StreamingExtension.Replay = function(channel, replayId) {
   }
   
   this.outgoing = function(message, callback) {
-    if (message.channel === '/meta/subscribe') {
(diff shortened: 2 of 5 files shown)


chandrasub:
Opened in error against the wrong repo — this change was intended for a private fork (PropelPLM/jsforce), not upstream. Apologies for the noise.

PR #1819, opened 2026-08-14
feat: Use URL API for OAuth2 endpoint construction @W-23807280@

k80bowman:
### What does this PR do?
Refactors OAuth2 endpoint URL construction to use the standard `URL` API instead of manual string manipulation (`split`/`join` and slash concatenation). This makes endpoint derivation more robust:

1. Derives `loginUrl` from `authzServiceUrl` via `new URL(authzServiceUrl).origin` rather than splitting on `/`.
2. Introduces an `appendPath` helper that safely joins the base URL and service path using the `URL` API, correctly handling trailing slashes.
3. Preserves existing behavior — service paths are percent-encoded consistently, already-encoded paths are not double-encoded, and the default `loginUrl` is used when none is provided.

Note: a scheme-less `loginUrl` (e.g. `test.salesforce.com`) now throws a `TypeError: Invalid URL`, matching standard `URL` parsing behavior.

### What issues does this PR fix or reference?
[@W-23807280@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002hCJ0nYAG/view)

### DIFF
--- a/src/oauth2.ts
+++ b/src/oauth2.ts
@@ -10,6 +10,13 @@ const defaultOAuth2Config = {
   loginUrl: 'https://login.salesforce.com',
 };
 
+const appendPath = (base: string, segment: string): string => {
+  const url = new URL(base);
+  const basePath = url.pathname.endsWith('/') ? url.pathname.slice(0, -1) : url.pathname;
+  url.pathname = `${basePath}/${segment}`;
+  return url.href;
+};
+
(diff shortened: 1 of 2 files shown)


jfeingold35 (approved):

jfeingold35:
QA steps: manually copy-pasted relevant code into CLI install, and verified:
- issue no longer reproducible
- non-malicious repos unaffected

PR #1824, opened 2026-08-18
W-23797589-fix: set allowH2:false to prevent GOAWAY crashes

soridalac:
### Summary

  - undici 8.x defaults to HTTP/2, causing unhandled SocketError crashes when Salesforce servers send GOAWAY frames during any sf command
  - Disable HTTP/2 by setting allowH2: false on the undici dispatcher for both direct and proxied connections

### Fix

  Force HTTP/1.1 by passing allowH2: false to the undici dispatcher, restoring the pre-v3.10.17 behavior where jsforce never used HTTP/2.

Cli issue related: https://github.com/forcedotcom/cli/issues/3622
@W-23797589@

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events';
 import { Duplex, Readable, Writable } from 'stream';
-import {fetch, errors, Response, RequestInit, ProxyAgent} from 'undici';
+import {fetch, errors, Response, RequestInit, ProxyAgent, Agent} from 'undici';
 import {
   createHttpRequestHandlerStreams,
   executeWithTimeout,
@@ -10,6 +10,9 @@ import {
 import { HttpRequest, HttpRequestOptions } from './types';
 import { getLogger } from './util/logger';
 import is from '@sindresorhus/is';
+
+// undici 8.x defaults to H2, causing crashes on server GOAWAY frames
+const noH2Agent = new Agent({ connect: { allowH2: false } });
 
 type CodedError = {
   code?: string;
@@ -40,7 +43,7 @@ async function startFetchRequest(
 ) {
   const logger = getLogger('fetch');
   const { httpProxy, followRedirect } = options;
-  const agent = httpProxy ? new ProxyAgent(httpProxy) : undefined;
+  const agent = httpProxy ? new ProxyAgent({ uri: httpProxy, requestTls: { allowH2: false } }) : noH2Agent;
   const { url, body, ...rrequest } = request;
   const controller = new AbortController();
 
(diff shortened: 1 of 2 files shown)


src/request.ts
```diff
@@ -40,7 +43,7 @@ async function startFetchRequest(
 ) {
   const logger = getLogger('fetch');
   const { httpProxy, followRedirect } = options;
-  const agent = httpProxy ? new ProxyAgent(httpProxy) : undefined;
+  const agent = httpProxy ? new ProxyAgent({ uri: httpProxy, requestTls: { allowH2: false } }) : noH2Agent;
```
WillieRuemmele: should this always be set, `requestTls: { allowH2: false } }`? no matter `httpProxy`... just instantiate a new `ProxyAgent(requestTls: { allowH2: false } })` in the `else` case of the ternary?

soridalac: I updated to fixed by setting allowH2: false on the global dispatcher instead, so the non-proxy path uses undefined (falls through to global). `ProxyAgent` still needs the explicit `requestTls: { allowH2: false }` because it has its own TLS config for the target connection. We can't use `ProxyAgent `for the non-proxy path — it requires a proxy URI and throws without one.

WillieRuemmele (approved):

WillieRuemmele:
## QA Results

Reproduced the GOAWAY crash locally using a test harness that spins up an H2 server which sends GOAWAY frames after every 3rd request, then fires concurrent requests through a shared undici dispatcher (mimicking how jsforce reuses connections).

### Without fix (H2 enabled):
```
[round 2/5] Firing 4 concurrent requests...
[server] Sending GOAWAY after request #6 (killing in-flight streams)
  [client] Request #5: FAILED — TypeError: fetch failed
  [client] Request #6: FAILED — TypeError: fetch failed
  [client] Request #7: FAILED — TypeError: fetch failed
  [client] Request #8: FAILED — TypeError: fetch failed

[result] 12/20 succeeded, 8/20 failed
```

### With fix (allowH2: false):
```
[result] 20/20 succeeded, 0/20 failed

✓ Fix verified — all requests succeeded with allowH2: false (HTTP/1.1 only).
```

The fix works — disabling H2 prevents GOAWAY from ever being a factor since HTTP/1.1 doesn't have that frame type.

---

## Requested change: avoid `setGlobalDispatcher`

The `setGlobalDispatcher` call at module load is a global side-effect — it disables H2 for **all** undici usage in the process, not just jsforce's requests. Any other library sharing the same process that intentionally uses H2 would be affected.

Since jsforce already centralizes fetch calls through `startFetchRequest`, the fix can be scoped to jsforce only by passing a per-request `dispatcher` instead:

```ts
// module-level agent, NOT global
const jsforceDispatcher = new Agent({ connect: { allowH2: false } });

// in startFetchRequest, pass it per-request:
const res = await fetch(url, { ...init, dispatcher: jsforceDispatcher });
```

This achieves the same result (H2 disabled for all jsforce connections) without the global side-effect. The `ProxyAgent` change is already correctly scoped per-request — just the `setGlobalDispatcher` line needs this treatment.

WillieRuemmele (changes requested):
see main comment

WillieRuemmele (approved):

